PR #2, opened 2022-05-27
Exploration

kylebarron:
- [x] External traits are cool
- [x] geozero is cool
- [ ] reconstruct series from output of geospatial algorithm. Haven't figured out yet how to construct an arrow array of `List[u8]` through the rust API. nested types don't feel as clearly documented in polars.
- [ ] Custom traits to construct geo types through geozero without building a `vec<u8>`


nmandery:
I think you can constuct a list types by creating a Series from a Vec<Series> using the NamedFrom trait in https://github.com/pola-rs/polars/blob/e21bdefc90280bdc2af59ab90e16dce08b93cb97/polars/polars-core/src/named_from.rs#L117 . I suppose there are more ways for this construction, but I am limited to a smartphone currently, so it is a bit hard to check.

kylebarron:
Oh sorry @nmandery I missed your comment before.

I think my issue was more with using nested types rather than the polars Series API specifically. For now I'm trying to get things working using WKB geometries, and then in the future can revisit using an Arrow-native geometry encoding.

Polars doesn't expose a `BinaryArray` type specifically; instead it interprets Arrow's `BinaryArray` type as a nested list of `u8`. For now I got things working by constructing the output array using `arrow2`'s `MutableBinaryArray` type directly.

PR #11, opened 2022-06-01
Bounds of geometry column

kylebarron:
This doesn't work because Polars doesn't support `FixedSizeListArray`
```
Error: InvalidOperation("Cannot create polars series from FixedSizeList(Field { name: \"item\", data_type: Float64, is_nullable: true, metadata: {} }, 4) type")
```

Regardless, I'm not sure the best way to be constructing these columns


kylebarron:
Closing as stale

PR #18, opened 2022-06-08
Adds convex_hull to geoseries and geodataframe

stuartlynn:
This PR adds a convex_hull function for GeoSeries and GeoDataFrames.

The function is implemented for the follow GeoSeries types
- MultiPoint 
- MultiPolygon
- Polygon
- LineString 
- MultiLineString

Added test for MultiPoint case

### DIFF
--- a/geopolars/src/geodataframe.rs
+++ b/geopolars/src/geodataframe.rs
@@ -3,11 +3,17 @@ use polars::prelude::{DataFrame, Result, Series};
 
 pub trait GeoDataFrame {
     fn centroid(&self) -> Result<Series>;
+    fn convex_hull(&self) -> Result<Series>;
 }
 
 impl GeoDataFrame for DataFrame {
     fn centroid(&self) -> Result<Series> {
         let geom_column = self.column("geometry")?;
         geom_column.centroid()
+    }
+
+    fn convex_hull(&self) -> Result<Series> {
+        let geom_column = self.column("geometry")?;
+        geom_column.convex_hull()
     }
 }
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -26,6 +26,11 @@ pub trait GeoSeries {
     /// geometry.
     fn envelope(&self) -> Result<Series>;
 
+    /// Returns a GeoSeries of geometries representing the convex hull of each geometry.
+    ///
+    /// The convex hull of a geometry is the smallest convex Polygon containing all the points in each geometr
+    fn convex_hull(&self) -> Result<Series>;
+
     /// Returns a GeoSeries of LinearRings representing the outer boundary of each polygon in the
     /// GeoSeries.
     ///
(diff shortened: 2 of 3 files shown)


stuartlynn:
Just getting my feet wet with this repo by adding this function. Follows the same pattern as envelope.

A few questions here 

1. I included a test in the GeoSeries as I didn't see tests set up anywhere else. Is this the structure for testing that we want to use in the crate?
2. I only tested against the MultiPoint case because it tests the actual execution of the function on the series. The actual testing of the convex_hull function happens in the geo crate so I don't see much point in adding tests for the other types. Does that make sense?
3. For functions like envelope and convex_hull, it would make sense to have a version that operates on the series as a whole. This would return the convex_hull / envelope for the combined collection of geometries in the series. If we where to add these functions, how would we want the API to be structured?  `df.convex_hull_all() ` for example?

geopolars/src/geoseries.rs
```diff
@@ -26,6 +26,11 @@ pub trait GeoSeries {
     /// geometry.
     fn envelope(&self) -> Result<Series>;
 
+    /// Returns a GeoSeries of geometries representing the convex hull of each geometry.
+    ///
+    /// The convex hull of a geometry is the smallest convex Polygon containing all the points in each geometr
+    fn convex_hull(&self) -> Result<Series>;
```
kylebarron, line 32: Nit: I was thinking of keeping this in alphabetical order. Not sure if that's important or just ocd 🤷‍♂️

geopolars/src/geodataframe.rs
```diff
@@ -3,11 +3,17 @@ use polars::prelude::{DataFrame, Result, Series};
...
+    fn convex_hull(&self) -> Result<Series>;
 }
 
 impl GeoDataFrame for DataFrame {
     fn centroid(&self) -> Result<Series> {
         let geom_column = self.column("geometry")?;
         geom_column.centroid()
     }
+
+    fn convex_hull(&self) -> Result<Series> {
```
kylebarron, line 15: I'm not sure what the right approach is here. This `GeoDataFrame` construct was just hacking; you can see the hard coded `"geometry"`. Ideally we'd be able to store dataframe-level metadata about a "primary geometry column" as I think geopandas does.

geopolars/src/geoseries.rs
stuartlynn: Happy to move it to the correct alphabetic location :-)

kylebarron:
> 1. I included a test in the GeoSeries as I didn't see tests set up anywhere else. Is this the structure for testing that we want to use in the crate?

Yes I hadn't gotten to tests yet 😅. In-file tests seems to be the standard for rust, so I'm 👍 for that. (Also note that CI isn't set up to run tests yet)

> 2. I only tested against the MultiPoint case because it tests the actual execution of the function on the series. The actual testing of the convex_hull function happens in the geo crate so I don't see much point in adding tests for the other types. Does that make sense?

Yeah. I think the actual test setup could still be discussed further.

> 3. For functions like envelope and convex_hull, it would make sense to have a version that operates on the series as a whole. This would return the convex_hull / envelope for the combined collection of geometries in the series. If we where to add these functions, how would we want the API to be structured?  `df.convex_hull_all() ` for example?

Good question. I'm not sure. Is there a similar entire-series function in the geopandas api?

geopolars/src/geodataframe.rs
stuartlynn: Humm, trying to think what functionality is available on a GeoPandas dataframe that isn't just calling down to the GeoSeries .

Some of the plotting functions for sure, where you need access to both the geometry column and a variable to color the geometries with. Some of the spatial join functions might naturally live at the DataFrame level as well?

I like the idea of having a primary geometry column like GeoPandas.

kylebarron: It seems like _most_ functions don't _need_ a primary geo column set. Like in pandas you have `pd.merge(left, right, left_on, right_on)`, and so you could do the same with a spatial join where you just define which column in each df you should merge on.

I'm not necessarily opposed to having a primary geo column but it seems like it's blocked anyways on polars exposing dataframe-level metadata (unless we just declare the primary geo column needs to have the name "geometry").

stuartlynn:
For the test setup it might be worth having a few convince functions to quickly construct GeoSeries from arrays of GeoType objects. Constructing the arrow arrays themselves is a little verbose. I suspect that could be useful functionality outside of the tests themselves.


I dont think that GeoPandas does have Series levels versions of convex_hull or envelope , I think the approatch most people use with it is to call unary_union on the Series and then the associated function so something like 

```python
series = GeoSeries(...)
series.unary_union.convex_hull()
```

geopolars/src/geodataframe.rs
stuartlynn: Yeah, I think that it's handy for convenience or when the DataFrame only has a single geometry column. 

Sorry the join example was more in thinking about the usefulness of GeoDataFrame construct in general rather than the idea of a primary geometry column.

kylebarron: This is fine for now and we can come back to what the geodataframe API should look like

PR #34, opened 2022-06-11
Adding 2 length methods to GeoSeries

### LINKED ISSUE #30: Implement: Length
Given a GeoSeries, this should return the length of each of the geometries in the series in the units of the series CRS.
For MultiPolygons it should return the length of the exterior 

Geopandas equivalent: [length](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.length.html)
Implemented in Geo crate : Yes , it implements [geodesic length](https://docs.rs/geo/latest/geo/algorithm/geodesic_length/index.html) for lines, so we would need to first calculate the exterior of a geometry then use this function to calculate the length

stuartlynn:
This adds the haversine, euclidean and geodesic length methods to the GeoSeries. 

For Polygons, MultiPolygons, Rects and Triangles, the methods calculate the total length of the exterior rings 

For Lines / MutliLines and LineStrings it returns the total length of the lines

For Points / MultiPoints it returns 0 as they technically have 0 length (this is what GeoPandas does as well)

It returns an error for GeometryCollections.

This should resolve #30

### DIFF
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -9,6 +9,12 @@ use geo::{Geometry, Point};
 use geozero::{CoordDimensions, ToWkb};
 use polars::prelude::{PolarsError, Result, Series};
 
+pub enum GeodesicLengthMethod {
+    Haversine,
+    Geodesic,
+    Vincenty,
+}
+
 pub trait GeoSeries {
     /// Returns a Series containing the area of each geometry in the GeoSeries expressed in the
     /// units of the CRS.
@@ -21,7 +27,7 @@ pub trait GeoSeries {
 
     /// Returns a GeoSeries of geometries representing the convex hull of each geometry.
     ///
-    /// The convex hull of a geometry is the smallest convex Polygon containing all the points in each geometr
+    /// The convex hull of a geometry is the smallest convex Polygon containing all the points in each geometry
     fn convex_hull(&self) -> Result<Series>;
 
     /// Returns a GeoSeries of geometries representing the envelope of each geometry.
@@ -51,7 +57,7 @@ pub trait GeoSeries {
     /// Not valid for Point or MultiPoint geometries. For Polygon it's the
     /// length of the exterior ring of the exterior ring of the Polygon and for MultiPolygon
     /// it returns the
-    fn geodesic_length(&self) -> Result<Series>;
+    fn geodesic_length(&self, method: GeodesicLengthMethod) -> Result<Series>;
 
     /// Returns the type ids of each geometry
     /// This mimics the pygeos implementation
@@ -67,14 +73,6 @@ pub trait GeoSeries {
     /// MULTIPOLYGON is 6
     /// GEOMETRYCOLLECTION is 7
     fn geom_type(&self) -> Result<Series>;
-
-    /// Returns a Series with the value of the geodesic length of each geometry
-    ///
-    /// Calculates the haversine length of each geometry in the series and returns it as a series.
-    /// Not valid for Point or MultiPoint geometries. For Polygon it's the
-    /// length of the exterior ring of the exterior ring of the Polygon and for MultiPolygon
-    /// it returns the
-    fn haversine_length(&self) -> Result<Series>;
 
     /// Returns a boolean Series with value True for empty geometries
     fn is_empty(&self) -> Result<Series>;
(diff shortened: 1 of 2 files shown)


geopolars/src/geoseries.rs
```diff
@@ -31,12 +31,28 @@ pub trait GeoSeries {
...
+    /// it returns the
+    fn euclidean_length(&self) -> Result<Series>;
+
     /// Returns a GeoSeries of LinearRings representing the outer boundary of each polygon in the
     /// GeoSeries.
     ///
     /// Applies to GeoSeries containing only Polygons. Returns `None` for other geometry types.
     fn exterior(&self) -> Result<Series>;
 
+    /// Returns a Series with the value of the geodesic length of each geometry
```
kylebarron, line 48: I don't really know: are geodesic and haversine lengths the same concept, just that geodesic is more accurate? I.e. euclidean length returns values in the same CRS as input, but geodesic and haversine both take as input spherical coordinates and return distances in meters?

Just curious because I'm wondering if maybe the API should be `geodesic_length(method='haversine')` (maybe just in the Python API, assuming that people using the rust api would be familiar with `geo`)

geopolars/src/geoseries.rs
```diff
@@ -150,6 +174,41 @@ impl GeoSeries for Series {
...
+                    .map(|poly| poly.exterior().euclidean_length())
+                    .sum()),
+                Geometry::GeometryCollection(_) => {
+                    Err(PolarsError::ComputeError(std::borrow::Cow::Borrowed(
+                        "Length methods are not implemented for geometry collection",
+                    )))
+                }
+                Geometry::Rect(rec) => Ok(rec.to_polygon().exterior().euclidean_length()),
+                Geometry::Triangle(triangle) => {
+                    Ok(triangle.to_polygon().exterior().euclidean_length())
```
kylebarron, line 202: I'm not sure whether we should generally include these or just stick with the standard geojson types

geopolars/src/geoseries.rs
```diff
@@ -193,6 +287,42 @@ impl GeoSeries for Series {
...
+                    Ok(multi_line_string.haversine_length())
+                }
+                Geometry::MultiPolygon(mutli_polygon) => Ok(mutli_polygon
+                    .iter()
+                    .map(|poly| poly.exterior().haversine_length())
+                    .sum()),
+                Geometry::GeometryCollection(_) => {
+                    Err(PolarsError::ComputeError(std::borrow::Cow::Borrowed(
+                        "Length methods are not implemented for geometry collection",
+                    )))
```
kylebarron, line 311: I guess in the future to support geometry collections it'll be easiest to make helper functions that can recurse into themselves.

kylebarron (approved):

kylebarron:
> this is what GeoPandas does as well

Maybe in the long term we should port geopandas tests here as well

geopolars/src/geoseries.rs
stuartlynn: Yeah I was tempted to go that way as well. They are as far as I know all just different methods to calculate the same thing.

I figure it doesn't hurt to have them all in there but we should perhaps create a wrapper  method both in rust and python that defaults to whatever one we think is the best to use.

I actually missed one of the implementations: Vincenty’s formula which I can add quickly.

geopolars/src/geoseries.rs
stuartlynn: I think that it makes sense to handle them if we can. I think they will rarely get used but it seems simple enough to convert them to polygons and then just treat them the same as polygons.

geopolars/src/geoseries.rs
stuartlynn: I am not 100% sure how geopandas handles this, if at all. It would make sense to recurse through them but I think for now, let's just  leave them unimplemented.

geopolars/src/geoseries.rs
kylebarron: Though if the geoarrow geometry encoding itself doesn't support these types, then these branches will never occur, right? I don't know if the `geozero` wkb parser ever creates `Rect` or `Triangle` objects

geopolars/src/geoseries.rs
stuartlynn: Refactored this have a single method that takes a GeodesicLengthMethod argument.

geopolars/src/geoseries.rs
stuartlynn: Fair. Will leave in just now and we can always refactor out when we need to

py-geopolars/src/lib.rs
```diff
@@ -22,9 +22,29 @@ fn convex_hull(series: &PyAny) -> PyResult<PyObject> {
...
+
+#[pyfunction]
+fn geodesic_length(series: &PyAny) -> PyResult<PyObject> {
+    let series = ffi::py_series_to_rust_series(series)?;
+    let out = series
+        .geodesic_length(GeodesicLengthMethod::Geodesic)
+        .map_err(|e| PyValueError::new_err(format!("Something went wrong: {:?}", e)))?;
+    ffi::rust_series_to_py_series(&out)
+}
+
```
stuartlynn, line 42: @kylebarron Not too familiar with py03 and how it does arguments to functions. Whats the best way to pass through an optional named parameter for method here? 

We want the function signature to be series.geodesic_length(method="Geodesic")  but have series.geodesic_length() to default to method = "Geodesic" right?

Do we pass it in as an Option<PyObject> and transom in to a string which we match on?

kylebarron: I think you could either do the default on the Rust side, by taking the input to be `Option<String>` or `Option<Method>` for some method enum. Or you could do the default on the Python side via https://pyo3.rs/v0.16.4/class.html#method-arguments (I haven't tried this myself).

Not sure which is preferable. Maybe good to look at py-polars to see how they do it.

geopolars/src/geoseries.rs
```diff
@@ -169,6 +226,123 @@ impl GeoSeries for Series {
...
+                }
+                (GeodesicLengthMethod::Vincenty, Geometry::MultiLineString(multi_line_string)) => {
+                    multi_line_string
+                        .vincenty_length()
+                        .map_err(map_vincenty_error)
+                }
+                (GeodesicLengthMethod::Haversine, Geometry::MultiPolygon(mutli_polygon)) => {
+                    Ok(mutli_polygon
+                        .iter()
+                        .map(|poly| poly.exterior().haversine_length())
```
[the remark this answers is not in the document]
stuartlynn: @kylebarron This works for now I think we can refactor later. If that's good then I think we are good to go on this PR

py-geopolars/src/lib.rs
```diff
@@ -31,11 +31,19 @@ fn euclidean_length(series: &PyAny) -> PyResult<PyObject> {
...
 }
 
-#[pyfunction]
-fn geodesic_length(series: &PyAny) -> PyResult<PyObject> {
+#[pyfunction(args="(method=\"geodesic\")")]
+fn geodesic_length(series: &PyAny, method: &str) -> PyResult<PyObject> {
     let series = ffi::py_series_to_rust_series(series)?;
+
+    let rust_method : GeodesicLengthMethod  = match method{
+        "geodesic" => Ok(GeodesicLengthMethod::Geodesic),
```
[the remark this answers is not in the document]
stuartlynn: Yeah I agree. Not sure what else to call it through. 

https://docs.rs/geo/latest/geo/algorithm/geodesic_length/trait.GeodesicLength.html

That's the details on the method. Perhaps we can call it ellipsoid? So 

`series.geodesic_length(method="ellipsoid")`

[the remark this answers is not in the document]
kylebarron: What about `method="karney"`? That seems to refer to the name/author of the algorithm, similar to Vincenty?

I assume Vincenty also does its measurements on an ellipsoid? And haversine is the one that assumes a sphere?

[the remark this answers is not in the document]
kylebarron: Or the rust `geodesic_length` could just be the default here, where `vincenty` and `haversine` are the optional methods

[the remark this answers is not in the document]
kylebarron: I just meant `geodesic_length(method=None)` would do the rust `geodesic_length` while `geodesic_length(method='haversine')` would do the rust `haversine_length`. 3 separate functions is probably fine for now too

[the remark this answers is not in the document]
kylebarron: Not sure which will end up clearer

kylebarron:
Thanks!

kylebarron:
(Also btw let me know if you'd prefer to merge your own PRs 🙂 )

PR #35, opened 2022-06-12
Spatial join

stuartlynn:
Opening this as a draft PR while I work on it. 

- [x] Add methods for generating a SpatialIndex from GeoSeries
- [x] Implement spatial join with different predicates using the index


geopolars/src/geoseries.rs
```diff
@@ -169,6 +172,23 @@ impl GeoSeries for Series {
...
+                ))
+            })?;
+            wkb_array.push(Some(wkb));
+        }
+        let array: BinaryArray<i32> = wkb_array.into();
+
+        let series = Series::try_from(("geometry", Arc::new(array) as ArrayRef)).unwrap();
+        Ok(series)
+    }
+
```
stuartlynn, line 191: @kylebarron added this convenience function to more easily build GeoSeries. Let me know if it looks good and we can probably use it a few other places in the codebase to clean up the code a little.

Tried to make it a TryFrom implementation but cant implement the Trait for Series as it's coming from the Polars crate

geopolars/src/spatial_index.rs
```diff
@@ -0,0 +1,216 @@
...
+            ))),
+        }
+    }
+}
+
+pub struct SpatialIndex {
+    pub r_tree: RTree<TreeNode>,
+}
+
+impl SpatialIndex {}
```
stuartlynn, line 120: Eventually going to turn this in to a Trait and implement it for SpatialIndex. Probably rename SpatialIndex to RTreeIndex or something similar

geopolars/src/geodataframe.rs
```diff
@@ -16,4 +29,194 @@ impl GeoDataFrame for DataFrame {
...
     }
+
+    fn spatial_join(
+        lhs: &DataFrame,
+        rhs: &DataFrame,
+        join_type: JoinType,
+        predicate: Predicate,
+        _lsuffix: Option<&str>,
+        _rsuffix: Option<&str>,
+    ) -> Result<DataFrame> {
```
stuartlynn, line 40: Ok so lots still left to do here (both implementing and tidying up the code) but this is an initial stab at the spatial join. The strategy is 

1. Construct the spatial indexes for the two DataFrames 
2. Use the indexes to get candidate pairs of geometries who's envelopes overlap 
3. For each pair actually perform the relevant geometry check, this depends on the types of both of the geometries and the indicated predicate. 
4. For each pair that we get a hit for, record their indexes (locations in the series)

This is where things get a little weird and I think we can figure out ways to improve what happens next.

5. We add a column to the left hand and right hand datasets which is just a series of the index of each entry 
6. We construct a DataFrame of the join index tuples we  constructed at the geometry checking phase.
7. We do the relevant join between the lhs DataFrame and the join DataFrame 
8. Do the relevant join between the rhs DataFrame and the join DataFrame 
9. Drop the temporary indexes

This works but it means generating a few extra columns temporarily. If we had access to polars "join with series"  (https://github.com/pola-rs/polars/blob/793fbf849400243f530696c9233eab791a1c0911/polars/polars-core/src/frame/hash_join/mod.rs#L663 ) methods, I think we could do it just with the tuples but those are private.

I dont think we want to implement out own join logic ( as the code for this in polars is pretty involved) So this kind of approach, where we generate the pairs of  things that need joined, and then use the existing join code is how we want to do it, but this implementation feels clunky.

Open to suggestions!

stuartlynn: @kylebarron would love your thoughts here !

kylebarron: > 1. Construct the spatial indexes for the two DataFrames

Maybe this is a dumb question, but do we need to construct the spatial indexes for _both_ dataframes? Naively I would've thought that you could generate an index for just one. I.e. make a spatial index for the right dataframe and then iterate over every row in the left dataframe, checking against the tree of the right dataframe.

I don't have good ideas about the rest yet, but I agree that I'd love to not reimplement the low-level join code.

stuartlynn: I think for points and polygons, the point based spatial index isn't necessary.

For joins that involve two polygons datasets, I think there is some benefit to being able to quickly exclude non-overlapping envelopes. You could do this on iteration for sure, this is just a quick way of doing it. 

This was also just the quickest way to get a set of candidates in code  and test out the join proportion of the code. We can potentially optimize it for different cases as we go.

We can also make the code not generate the spatial indexes and instead just take as arguments. We could then use whatever indexes we have or manually iterate with one index, or manually iterate with no indexed depending on what's passed in.

geopolars/src/spatial_index.rs
```diff
@@ -0,0 +1,437 @@
...
+        PolarsError::ComputeError(std::borrow::Cow::Borrowed(
+            "Failed to generate the spatial index for the left dataframe",
+        ))
+    })?;
+
+    let spatial_index_right: SpatialIndex = rhs_geometry.try_into().map_err(|_| {
+        PolarsError::ComputeError(std::borrow::Cow::Borrowed(
+            "Failed to generate the spatial index for the right dataframe",
+        ))
+    })?;
```
stuartlynn, line 52: Not 100% sure how to optionally construct the index if none is passed. I tried 

```rust 
let spatial_index_left: &SpatialIndex = if let Some(index_ref) = options.l_index{
    index_ref
}
else{
    let new_index = lhs_geometry.try_into().map_err(|_| {
        PolarsError::ComputeError(std::borrow::Cow::Borrowed(
            "Failed to generate the spatial index for the left dataframe",
        ))?;
   &new_index
}
```

but this fails because new_index obviously goes out of scope at the end of the if let statement.

Not had to do this in rust before so not 100% sure of the idiomatic rust way of doing this. It seems like it should be possible to tell rust to hold on to new_index till the end of the function, but I am not 100% sure how to implement it (guessing some sort of lifetime?) @kylebarron thoughts?

stuartlynn: Made these Arcs which helps resolve that problem and also makes it easier to share indexes in the future between threads etc.

stuartlynn:
Going to merge this in and we can tweak as we go!

geopolars/src/spatial_index.rs
```diff
@@ -0,0 +1,621 @@
...
+    let lhs_join_series: Vec<u64> = left_series.iter().map(|i| *i as u64).collect();
+    let rhs_join_series: Vec<u64> = right_series.iter().map(|i| *i as u64).collect();
+
+    let lhs_join_series = Series::new("lhs_join", lhs_join_series);
+    let rhs_join_series = Series::new("rhs_join", rhs_join_series);
+
+    let join_df: DataFrame = DataFrame::new(vec![lhs_join_series, rhs_join_series])?;
+
+    let mut lhs_with_index = lhs.hstack(&[lhs_index])?;
+    let mut rhs_with_index = rhs.hstack(&[rhs_index])?;
```
kylebarron, line 159: What happens here if the left df already has a column named `lhs_join` or the right df already has a column `rhs_join`?

geopolars/src/spatial_index.rs
```diff
@@ -0,0 +1,621 @@
...
+            Ok(result)
+        }
+        JoinType::Left => {
+            let join_one = lhs_with_index.left_join(&join_df, ["lhs_index"], ["lhs_join"])?;
+            let join_two = join_one.left_join(&rhs_with_index, ["rhs_join"], ["rhs_index"])?;
+            let result = join_two.drop("lhs_index")?.drop("rhs_join")?;
+            Ok(result)
+        }
+        _ => Err(PolarsError::ComputeError(std::borrow::Cow::Borrowed(
+            "Failed to generate the spatial index for the left dataframe",
```
kylebarron, line 194: Is this the right error text? This is separate from the spatial index, right?

geopolars/src/spatial_index.rs
kylebarron: > because new_index obviously goes out of scope at the end of the if let statement

Hmm. So the issue here is we want to handle both when we don't own the memory (spatial index is passed in) or when we do own the memory (we generate it)? I don't know what the idiomatic rust way of doing that would be.

> Made these Arcs which helps resolve that problem

That seems fine to me. We can always optimize further in the future if needed

geopolars/src/util.rs
```diff
@@ -13,3 +13,18 @@ pub fn iter_geom(series: &Series) -> impl Iterator<Item = Geometry<f64>> + '_ {
         Wkb(vec).to_geo().expect("unable to convert to geo")
     })
 }
+
+pub fn geom_at_index(series: &Series, index: usize) -> Result<Geometry<f64>> {
+    let chunks = series.list().expect("series was not a list type");
+    let row = chunks.into_iter().nth(index);
```
kylebarron, line 19: I learned recently that `nth` iterates through (and discards) every element up to `n`. So it's `O(n)` not `O(1)`; and if we iterated over the series calling `geom_at_index` it would end up being `O(n^2)` I think.

I'm sure there _must_ be some polars method that accesses a value at an index in constant time. I looked through the [`Series`](https://pola-rs.github.io/polars/polars/series/struct.Series.html) and [`ChunkedArray`](https://pola-rs.github.io/polars/polars/chunked_array/struct.ChunkedArray.html) APIs quickly, though didn't immediately find the right method. I'm just a little worried that this `into_iter().nth()` is a bottleneck.

Alzavio:
Just curious if this was ever completed / documented.

kylebarron:
This old PR isn't relevant to the current state of GeoPolars, which is expected to be built on top of https://github.com/geoarrow/geoarrow-rs

jreniel:
Hi @kylebarron,
Thanks for the pointer to geoarrow-rs! Just to clarify, is spatial join functionality (similar to gpd.sjoin) planned for the geoarrow-rs-based GeoPolars? I see [geoarrow/geoarrow-rs#652](https://github.com/geoarrow/geoarrow-rs/issues/652) tracks this, but it looks like it hasn't landed yet. Is that the right place to follow progress?                                                                                                                                                                                                                                                
I am building a ton of Rust based tools that require spatial joins and I have been following the progress on all the geo crates since early. I was happy to hear a couple weeks ago that things started to move forwards again. Thanks for the hard work.

kylebarron:
Ideally we'll have spatial join functionality in geoarrow-rs, yes. I think we want GeoPolars to be as small a shim over geoarrow-rs as possible.

jreniel:
Thanks! Looking forwards to trying out the new developments. Good luck!

PR #36, opened 2022-06-14
Create GeoSeries and GeoDataFrame subclasses from polars in Python

### LINKED ISSUE #33: Subclass from `polars.DataFrame` in Python
Similarly to how `geopandas.GeoDataFrame` subclasses from `pandas.DataFrame`, I think it would be clean if `geopolars.GeoDataFrame` subclassed from `polars.DataFrame`. Then all the methods from polars are still available.

In https://github.com/kylebarron/geopolars/pull/4 I originally took a stab at this, but ran into a mess because when subclassing _from within Rust_, it appears that you need to additionally need to compile the library you're wrapping. We really don't want to compile polars' Python API.

Instead, we should explore doing this subclassing _from within Python_. I'm not 100% on how `py-polars` is set up, but it has both [Rust code](https://github.com/pola-rs/polars/tree/4e83b4012e06c84581ce9fe90d0b664cb4a054d4/py-polars/polars) _and_ [Python code](https://github.com/pola-rs/polars/tree/4e83b4012e06c84581ce9fe90d0b664cb4a054d4/py-polars/polars) within the `py-polars` directory. So it seems like pyo3/maturin have a way to distribute both rust code and python code together. (It looks like in the polars case, [`polars.polars` is the entrypoint to the rust bindings](https://github.com/pola-rs/polars/blob/4e83b4012e06c84581ce9fe90d0b664cb4a054d4/py-polars/polars/internals/frame.py#L59)).

When the subclassing happens from within Python, it also means that `polars` and `geopolars` could be installed separately and subclassed at runtime, and so we shouldn't have issues with version conflicts that could happen if both were compiled against different versions of polars.

kylebarron:
Closes #33.

The goal of this is to create a similar API to geopandas, where it subclasses `Series` and `DataFrame`, assigning methods directly onto those classes.

Questions:

- What's the best way to return an instance of `GeoDataFrame` and `GeoSeries`? Ideally, the actual data representation contained in each of these won't be different; it'll just be the python object that additionally has more methods. The pure functions in the python-rust bindings currently return a bare polars Series, so we could:
    - Just wrap the return object in a call to `GeoSeries(...)` in Python, i.e.

        ```py
        def centroid(self) -> GeoSeries:
            return GeoSeries(geopolars.centroid(self))
        ```

    - Try to make this conversion automatically from the rust side? The current rust bindings call `polars.from_arrow` to reconstruct a polars series: 
https://github.com/kylebarron/geopolars/blob/71cd2027617324a15c609450ec90fafe3b90213c/py-geopolars/src/ffi.rs#L92 Could we just create a `geopolars.from_arrow` function that returns a `GeoSeries`/`GeoDataFrame`? I think this might actually be easier, as long as we don't run into any weird circular dependency issues.


py-geopolars/src/ffi.rs
```diff
@@ -1,5 +1,5 @@
 use arrow::{array::ArrayRef, ffi};
-use polars::prelude::*;
+use polars::prelude::{ArrowField, Series};
```
kylebarron, line 2: This is just a nit that I don't love imports into the global namespace. Especially for things like `ArrowField` that could have come from `arrow2`

kylebarron:
I don't know why there are obscure [`cargo metadata` errors](https://github.com/kylebarron/geopolars/runs/6874508314?check_suite_focus=true#step:5:11) (only on CI) when adding a python package to the top-level workspace. (note that the error is about `geopolars/geopolars/py-geopolars`, when that folder doesn't exist)

But the python libs also aren't in the polars workspace: https://github.com/pola-rs/polars/blob/4154c6af37cc366868289059f70d5146d72d5848/Cargo.toml#L2-L13

stuartlynn:
> Closes #33.
> 
> The goal of this is to create a similar API to geopandas, where it subclasses `Series` and `DataFrame`, assigning methods directly onto those classes.
> 
> Questions:
> 
> * What's the best way to return an instance of `GeoDataFrame` and `GeoSeries`? Ideally, the actual data representation contained in each of these won't be different; it'll just be the python object that additionally has more methods. The pure functions in the python-rust bindings currently return a bare polars Series, so we could:
>   
>   * Just wrap the return object in a call to `GeoSeries(...)` in Python, i.e.
>     ```python
>     def centroid(self) -> GeoSeries:
>         return GeoSeries(geopolars.centroid(self))
>     ```
>   * Try to make this conversion automatically from the rust side? The current rust bindings call `polars.from_arrow` to reconstruct a polars series:
>     https://github.com/kylebarron/geopolars/blob/71cd2027617324a15c609450ec90fafe3b90213c/py-geopolars/src/ffi.rs#L92
>      Could we just create a `geopolars.from_arrow` function that returns a `GeoSeries`/`GeoDataFrame`? I think this might actually be easier, as long as we don't run into any weird circular dependency issues.

I think my preference here would be to make this as automatic as we can, esp for future contributors who might not catch the need to return the series wrapped as a GeoSeries. 

One small concern though is that we will have times when the API is actually returning a Series, rather than a GeoSeries. For example the method that calculates the area. Should we have two functions then? One for rust_series_to_py_series and one rust_series_to_py_geoseries ? We presumably wouldn't want the python series to have the Geo functions if it's a regular series? 

Perhaps instead of having two functions we can make the rust_series_to_py_series inspect the series it is passed to see if it's a geoseries and conditionally wrap the series in either a vanilla Series or a GeoSeries

py-geopolars/src/ffi.rs
kylebarron: oh no you didn't add that! I was just explaining the diff 🙂

kylebarron:
I should have some time tomorrow afternoon to finish up this PR

kylebarron:
I added a `from_arrow` function in `convert.py` (mirroring the layout of `py-polars`) and added  `rust_series_to_py_geoseries` in addition to `rust_series_to_py_series` in ffi.rs.

kylebarron:
I added a `from_arrow` function in `convert.py` (mirroring the layout of `py-polars`) and added  `rust_series_to_py_geoseries` in addition to `rust_series_to_py_series` in ffi.rs.

PR #39, opened 2022-07-01
Adding affine transform methods to GeoSeries

stuartlynn:
Adds affine transform methods from Geo to the GeoSeries 

Mehods added 
- translate 
- rotate
- skew
- scale
- affine_transform

Notes: 

- Added a Transform Origin to allow easy specification of the origin of a transform for those that accept one. 
- Currently the arguments of a transform are specified only by manual values. It would be great if you could specify another Series for these as well so that each geometry would be transformed according to the parameters in that series.
- The current implementation does not work in-place, we might want to have another set of functions to do the transforms in-place, updating the geometry column with the new geoms 
- Currently there is no method to chain together transforms, instead I would advise people compose the transform using methods from the geo crate and then apply that transform with the affine_transform method.

### DIFF
--- a/geopolars/Cargo.toml
+++ b/geopolars/Cargo.toml
@@ -6,8 +6,9 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-arrow2 = {version = "0.11.2", features = ["io_ipc"]}
-geo = "0.20.1"
-geozero = {version = "0.9.4", features = ["with-wkb"]}
-polars = {version = "0.21.1", features = ["ipc", "dtype-u8", "dtype-i8"]}
+arrow2 = { version = "0.11.2", features = ["io_ipc"] }
+geozero = { version = "0.9.4", features = ["with-wkb"] }
 rstar = "0.9.3"
+geo = { git = "https://github.com/georust/geo.git", rev = '578e213875915e1f895892487b5a36ca0d91fac3' }
+polars = { version = "0.21.1", features = ["ipc", "dtype-u8", "dtype-i8"] }
+geo-types = { version = "0.7.6", features = ["approx", "use-rstar_0_9"] }
--- a/geopolars/Cargo.toml
+++ b/geopolars/Cargo.toml
@@ -11,4 +11,3 @@ geozero = { version = "0.9.4", features = ["with-wkb"] }
 rstar = "0.9.3"
 geo = { git = "https://github.com/georust/geo.git", rev = '578e213875915e1f895892487b5a36ca0d91fac3' }
 polars = { version = "0.21.1", features = ["ipc", "dtype-u8", "dtype-i8"] }
-geo-types = { version = "0.7.6", features = ["approx", "use-rstar_0_9"] }
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -130,27 +130,6 @@ pub trait GeoSeries {
     fn y(&self) -> Result<Series>;
 }
 
-// struct Geom(geo::Geometry);
-
-// impl FromIterator<Geom> for Series{
-//     fn from_iter<T: IntoIterator<Item = Geom>>(iter: T) -> Self {
-//         let mut wkb_array = MutableBinaryArray::<i32>::new();
-
-//         for geom in iter{
-//             let wkb = geom.to_wkb(CoordDimensions::xy()).map_err(|_| {
-//                 PolarsError::ComputeError(std::borrow::Cow::Borrowed(
-//                     "Failed to convert geom vec to GeoSeries",
-//                 ))
-//             }).unwrap();
-//             wkb_array.push(Some(wkb));
-//         }
-//         let array: BinaryArray<i32> = wkb_array.into();
-
-//         let series = Series::try_from(("geometry", Arc::new(array) as ArrayRef)).unwrap();
-//         series
-//     }
-// }
-
 impl GeoSeries for Series {
     fn affine_transform(&self, matrix: impl Into<AffineTransform<f64>>) -> Result<Series> {
         let transform: AffineTransform<f64> = matrix.into();
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -19,13 +19,14 @@ pub enum GeodesicLengthMethod {
 
 /// Used to express the origin for a given transform. Can be specified either be with reference to
 /// the geometry being transformed (Centroid, Center) or some arbitrary point.
+///
+/// - Centroid: Use the centriod of each geometry in the series as the transform origin.
+/// - Center: Use the center of each geometry in the series as the transform origin. The center is
+///   defined as the center of the bounding box of the geometry
+/// - Point: Define a single point to transform each geometry in the series about.
 pub enum TransformOrigin {
-    /// Use the centriod of each geometry in the series as the transform origin.
     Centroid,
-    /// Use the center of each geometry in the series as the transform origin. The center is
-    /// defined as the center of the bounding box of the geometry
     Center,
-    /// Define a single point to transform each geometry in the series about.
     Point(Point),
 }
 
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -5,9 +5,11 @@ use arrow2::array::{
     ArrayRef, BinaryArray, BooleanArray, MutableBinaryArray, MutableBooleanArray,
     MutablePrimitiveArray, PrimitiveArray,
 };
-use geo::{Geometry, Point};
+use geo::algorithm::affine_ops::AffineTransform;
+use geo::{map_coords::MapCoords, Geometry, Point};
 use geozero::{CoordDimensions, ToWkb};
 use polars::prelude::{PolarsError, Result, Series};
+use std::convert::Into;
 
 pub enum GeodesicLengthMethod {
     Haversine,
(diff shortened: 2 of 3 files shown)


urschrei:
Great! In-place methods are imminent btw: https://github.com/georust/geo/pull/872

geopolars/Cargo.toml
```diff
@@ -6,8 +6,9 @@ edition = "2021"
...
-arrow2 = {version = "0.11.2", features = ["io_ipc"]}
-geo = "0.20.1"
-geozero = {version = "0.9.4", features = ["with-wkb"]}
-polars = {version = "0.21.1", features = ["ipc", "dtype-u8", "dtype-i8"]}
+arrow2 = { version = "0.11.2", features = ["io_ipc"] }
+geozero = { version = "0.9.4", features = ["with-wkb"] }
 rstar = "0.9.3"
+geo = { git = "https://github.com/georust/geo.git", rev = '578e213875915e1f895892487b5a36ca0d91fac3' }
+polars = { version = "0.21.1", features = ["ipc", "dtype-u8", "dtype-i8"] }
+geo-types = { version = "0.7.6", features = ["approx", "use-rstar_0_9"] }
```
kylebarron, line 14: Is a direct dependency on `geo-types` the standard way to include `rstar`?

geopolars/src/geoseries.rs
```diff
@@ -87,6 +98,15 @@ pub trait GeoSeries {
     /// implicitly closed by copying the first tuple to the last index.
     fn is_ring(&self) -> Result<Series>;
 
+    /// Returns a GeoSeries with each of teh geometries roatered by a fixed x and y ammount arround
```
kylebarron, line 101:
```suggestion
    /// Returns a GeoSeries with each of the geometries rotated by a fixed x and y amount around
```

geopolars/src/geoseries.rs
```diff
@@ -87,6 +98,15 @@ pub trait GeoSeries {
...
+    /// some origin.
+
+    fn rotate(&self, angle: f64, origin: TransformOrigin) -> Result<Series>;
+
+    /// Returns a GeoSeries with each of the geometries skewd by a fixed x and y amount around a
+    /// given origin
+    fn scale(&self, xfact: f64, yfact: f64, origin: TransformOrigin) -> Result<Series>;
```
kylebarron, line 108: This has the same docstring as `skew` below.

geopolars/src/geoseries.rs
```diff
@@ -96,6 +116,13 @@ pub trait GeoSeries {
     /// https://docs.rs/geo/latest/geo/algorithm/simplify/trait.Simplify.html for details
     fn simplify(&self, tolerance: f64) -> Result<Series>;
 
+    /// Returns a GeoSeries with each of the geometries skewd by a fixed x and y amount around a
```
kylebarron, line 119:
```suggestion
    /// Returns a GeoSeries with each of the geometries skewed by a fixed x and y amount around a
```

geopolars/src/geoseries.rs
```diff
@@ -96,6 +116,13 @@ pub trait GeoSeries {
...
+    /// given origin
+    fn skew(&self, xs: f64, ys: f64, origin: TransformOrigin) -> Result<Series>;
```
kylebarron, line 121: Is this the same operation as geopandas' [`GeoSeries.skew`](https://geopandas.org/en/stable/docs/user_guide/geometric_manipulations.html#GeoSeries.skew)? Just wondering because their api takes in an `angle` as argument.

geopolars/src/geoseries.rs
```diff
@@ -87,6 +98,15 @@ pub trait GeoSeries {
     /// implicitly closed by copying the first tuple to the last index.
     fn is_ring(&self) -> Result<Series>;
 
+    /// Returns a GeoSeries with each of teh geometries roatered by a fixed x and y ammount arround
+    /// some origin.
+
+    fn rotate(&self, angle: f64, origin: TransformOrigin) -> Result<Series>;
```
kylebarron, line 104: It would be nice to include in the docstring whether `angle` is expected to be in degrees/radians.

geopolars/src/geoseries.rs
```diff
@@ -5,17 +5,28 @@ use arrow2::array::{
...
 }
 
+pub enum TransformOrigin {
+    Centroid,
+    Center,
+    Point(Point),
+}
+
 pub trait GeoSeries {
+    /// Apply an affine transform to the geoseries and return a geoseries of the tranformed geometries fn affine_transform(&self, matrix: impl Into<AffineTransform<f64>>  ) -> Result<Series>;
```
kylebarron, line 27: Is the function signature in the comment intentional?

geopolars/src/geoseries.rs
```diff
@@ -104,6 +131,15 @@ pub trait GeoSeries {
 }
 
 impl GeoSeries for Series {
+    fn affine_transform(&self, matrix: impl Into<AffineTransform<f64>>) -> Result<Series> {
+        let transform: AffineTransform<f64> = matrix.into();
+        let output_vec: Vec<geo::Geometry> = iter_geom(self)
```
kylebarron, line 136: Should we be consistent on using either `geo::Geometry` or bare `Geometry`? It looks like below you use mainly `Geometry`.

geopolars/src/geoseries.rs
```diff
@@ -5,17 +5,28 @@ use arrow2::array::{
...
 
 pub enum GeodesicLengthMethod {
     Haversine,
     Geodesic,
     Vincenty,
 }
 
+pub enum TransformOrigin {
+    Centroid,
+    Center,
```
kylebarron, line 22: It might be nice to have docstrings on these enum values. I wouldn't understand at a glance what the difference is between `Centroid` and `Center`. 

It looks like `Center` means "take the bounding rectangle and then take the center/centroid of that rectangle"? Would something like `BoundingRectCenter` be clearer as the enum value?

geopolars/src/geoseries.rs
```diff
@@ -550,6 +686,139 @@ mod tests {
         assert_eq!(result, correct, "Should get the correct convex hull");
     }
 
+    #[test]
+    fn skew() {
+        let geo_series = Series::from_geom_vec(&[Geometry::Polygon(polygon!(
```
kylebarron, line 691: Not that familiar with these macros, but does `polygon!` not create a `Geometry::Polygon` object already? You still have to wrap the output of `polygon!` with `Geometry::Polygon`?

geopolars/src/geoseries.rs
```diff
@@ -550,6 +686,139 @@ mod tests {
...
+        (x:0.008727532464108793,y:0.9825396152541261),
+        (x:1.008727532464109, y:1.0174603847458739),
+        (x:0.9912724675358912, y:0.017460384745873865)
+        ));
+
+        let skewed_series = geo_series.skew(1.0, 2.0, TransformOrigin::Center);
+        assert!(skewed_series.is_ok(), "To get a series back");
+
+        let geom = iter_geom(&skewed_series.unwrap()).next().unwrap();
+        assert_eq!(geom, result, "the swkfddfg");
```
kylebarron, line 710: More descriptive assertion text?

kylebarron (approved):
Some minor comments but in general everything looks great! Thanks!

kylebarron:
> In-place methods are imminent btw

I'm curious how much in-place methods would help us. Right now we're doing several copies for every geometry operation: WKB Arrow buffer -> `geo` struct -> algorithm -> `geo` struct -> WKB Arrow buffer. In-place operations would remove the `algorithm -> geo` copy, but I'm guessing that's a small portion of the overhead we currently have

geopolars/Cargo.toml
stuartlynn: It's not for the r-star dep it's for the affine transform support that just landed. Once that gets released we can move back to the a versioned number. Can prob take the r-star part out.

geopolars/src/geoseries.rs
stuartlynn: Nope that's en errant paste or something. Will remove

geopolars/src/geoseries.rs
stuartlynn: Again think this is an unfortunate past error :-)

geopolars/src/geoseries.rs
stuartlynn: Creates a geo-types polygon, not a Geometry::Polygon as far as I can tell. Switched these up to using an .into conversion though rather than wrapping them. Looks a little cleaner

geopolars/src/geoseries.rs
stuartlynn: yup changed this to use just Geometry

geopolars/src/geoseries.rs
stuartlynn: Added these

stuartlynn:
Yeah fair.  Think we can leave these just now as copy only and revisit when we are closer to having the trait based and arrow native formats

geopolars/src/geoseries.rs
stuartlynn: The angle is expressed along each axis with the xs and ys directions. Have added some docs to clarify

geopolars/Cargo.toml
stuartlynn: Actually I was able to remove this just fine without issue

geopolars/src/geoseries.rs
kylebarron: Ah I see; I forgot there was a difference

PR #54, opened 2022-07-14
[Python]: Build wheels for mac arm64

kylebarron:
Just curious to see if it works out of the box. Otherwise will probably go to the backburner.


kylebarron:
Closing in favor of #57 . Seems like this approach _should_ work as well. Maybe it's a bug in `maturin` but I'm not going to take the time to delve into it.

PR #57, opened 2022-07-15
Python: Build universal2 wheels for mac

kylebarron:


kylebarron:
Looks like the last commit worked in that 3.7 x86_64 still builds... but sadly it seems there's no caching between the x86_64 and the `universal2` builds. So now the mac wheels take forever, but at least this will only get run on releases 🤷‍♂️ 

```
✓ cp37-macosx_x86_64 finished in 1263.54s
✓ cp38-macosx_universal2 finished in 1196.70s
✓ cp39-macosx_universal2 finished in 64.20s
✓ cp310-macosx_universal2 finished in 59.84s
```

PR #62, opened 2022-07-15
Reprojection via proj

kylebarron:
### Change list

- Proj is added under a feature flag named `proj`
- `to_crs` method added to the trait and implementation
- Includes both `&from` and `&to` because we don't have a way to store ancillary metadata on the polars objects yet. Hopefully this will be alleviated in the future if polars supports extension types
- So this doesn't handle metadata _tracking_ at this point

### DIFF
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use crate::util::iter_geom;
 use geo::algorithm::affine_ops::AffineTransform;
-use geo::{map_coords::MapCoords, Coordinate, Geometry, Point};
+use geo::{map_coords::MapCoords, Geometry, Point};
 use geozero::{CoordDimensions, ToWkb};
 use polars::export::arrow::array::{
     ArrayRef, BinaryArray, BooleanArray, MutableBinaryArray, MutableBooleanArray,
@@ -631,17 +631,18 @@ impl GeoSeries for Series {
 
     #[cfg(feature = "proj")]
     fn to_crs(&self, from: &str, to: &str) -> Result<Series> {
-        use proj::{Proj, ProjError};
+        use proj::{Proj, Transform};
 
-        let proj_transformer = Proj::new_known_crs(from, to, None).unwrap();
-        let transform = |c: Coordinate<f64>| -> std::result::Result<_, ProjError> {
-            // proj can accept Point, Coordinate, Tuple, and array values, returning a Result
-            let shifted = proj_transformer.convert(c)?;
-            Ok(shifted)
-        };
-
+        let proj = Proj::new_known_crs(from, to, None).unwrap();
         let output_vec: Vec<Geometry> = iter_geom(self)
-            .map(|geom| geom.map_coords(|coord| transform(coord).unwrap()))
+            .map(|mut geom| {
+                // geom.tranform modifies `geom` in place.
+                // Note that this doesn't modify the _original series_ because iter_geom makes a
+                // copy
+                // https://docs.rs/proj/latest/proj/#integration-with-geo-types
+                geom.transform(&proj).unwrap();
+                geom
+            })
             .collect();
 
         Series::from_geom_vec(&output_vec)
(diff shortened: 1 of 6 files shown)


geopolars/src/geoseries.rs
```diff
@@ -624,6 +629,25 @@ impl GeoSeries for Series {
         }
     }
 
+    #[cfg(feature = "proj")]
+    fn to_crs(&self, from: &str, to: &str) -> Result<Series> {
+        use proj::{Proj, Transform};
+
+        let proj = Proj::new_known_crs(from, to, None).unwrap();
```
stuartlynn, line 636: This seems good for just now but it might be good to be able to support known projections and proj string definitions.

 We could perhaps implement an enum type 

```rust
enum ProjType {
    Known(string),
    Definition(string)
}
```

and implement `From<String> for ProjType ` in which we check to see if the string is of format "Authority:Code" in which case we return the Known variant and otherwise the Definition variant.

That would allow us to handle both cases while still allowing the user to pass in a simple string. Or we could even do without the enum and just use a regex to inspect the string and decide that way?

geopolars/src/geoseries.rs
```diff
@@ -624,6 +629,25 @@ impl GeoSeries for Series {
...
+        let output_vec: Vec<Geometry> = iter_geom(self)
+            .map(|mut geom| {
+                // geom.tranform modifies `geom` in place.
+                // Note that this doesn't modify the _original series_ because iter_geom makes a
+                // copy
+                // https://docs.rs/proj/latest/proj/#integration-with-geo-types
+                geom.transform(&proj).unwrap();
```
stuartlynn, line 643: Do we want to handle this error? Also more generally, if  an operation on a member of a Series fails, do we want the entire operation to fail ? I have wondered about that for a few different methods I implemented

stuartlynn (commented):
This looks great! Just a few minor points inline.

I think even if we cant store the crs metadata for the columns, this is still a hugely useful feature to have

geopolars/src/geoseries.rs
kylebarron: Yeah we definitely want to handle the error. I'm planning to come back and do a general overhaul to try and remove as many `unwrap`s as possible.

> Also more generally, if an operation on a member of a Series fails, do we want the entire operation to fail 

I lean towards yes, at least to start with. I think that mimics geopandas operations that fail...?

geopolars/src/geoseries.rs
kylebarron: I'm not sure I follow. Is your critique coming from the `known_crs` name?

This is my first time using the proj crate, but it looks like there are two ways to create a `Proj` object:
- Via [`Proj::new`](https://docs.rs/proj/latest/proj/struct.Proj.html#method.new), which uses a PROJ string definition like `"+proj=merc +lat_ts=56.5 +ellps=GRS80"` (and only takes that one argument)
- Via [`Proj::new_known_crs`](https://docs.rs/proj/latest/proj/struct.Proj.html#method.new_known_crs), which takes _both_ a `from` and a `to` argument, and takes a variety of string inputs:
    - an `"AUTHORITY:CODE"`, like `"EPSG:25832"`.
    - a PROJ string, like `"+proj=longlat +datum=WGS84"`
    - the name of a CRS as found in the PROJ database, e.g `"WGS84"`, `"NAD27"`, etc.
    - more generally, any string accepted by [new()](https://docs.rs/proj/latest/proj/struct.Proj.html#method.new)

It seems like here we'd always want the latter case? Where we want the high-level API to convert from one CRS to another?

The current behavior offloads the responsibility for passing in a valid PROJ CRS argument to the caller. I really don't want to do something like regex, because PROJ will do its own validation anyways. We should just be better about error handling (i.e. not have `unwrap`s like here)

geopolars/src/geoseries.rs
stuartlynn: Yeah I think it makes sense. There might be cases where returning a None might make sense. If the operation hasn't failed per-say but just isn't defined for a given geometry but yeah for now, it should be an explicit failure for the entire call I think

PR #64, opened 2022-07-15
Python: crs management in GeoSeries and GeoDataFrame

kylebarron:
### Change list

- Add a `crs` attribute to the Python `geoseries` subclass
- Maintain `crs` attribute when a method returns a new `GeoSeries`

This feels like a pretty hacky implementation, but it's a start.


stuartlynn (approved):
Looks good to me. Shame we can't do the CRS storage on the rust side but this is a nice way to support in python for nwo

kylebarron:
todo:

- Update `from_arrow` to check for geoarrow schema metadata
- tests to make sure functions are returning crs information correctly?
- Quick check to make sure all underlying rust geopolars functions are passing the crs correctly (leaving true tests for every function to another pr)

kylebarron:
Setting this milestone to 0.2 to wait until we have working proj wheels, see #130

kylebarron:
Closing, this is stale, and we can now manage CRS inside of the Polars extension type support

PR #65, opened 2022-07-15
Wasm directory scaffolding

kylebarron:
Basic scaffolding to create a new wasm binding.

It seems like it would be ideal if we could manage to have the wasm code in this folder, and just override `polars` (and maybe other dependencies) using [`patch.crates-io`](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section).

Then nothing in `geopolars` (the core rust module) needs to know about wasm; it just needs to stay general.


wasm-geopolars/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
```
kylebarron, line 1: This is mostly the default output from `wasm-pack new wasm-geopolars`

wasm-geopolars/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3.13"
+
+[profile.release]
+# Tell `rustc` to optimize for small code size.
+opt-level = "s"
+
+[patch.crates-io]
+polars = { git = "https://github.com/stuartlynn/polars.git", version = "0.22.6", branch = 'wasm_rebase', default-features = false, features = [
```
kylebarron, line 40: Ideally we'd be able to get this to compile _only_ by overriding dependencies here in the `patch` section

kylebarron:
@stuartlynn feel free to push to this branch or to create your own.

Ideally, it should compile correctly with just `wasm-pack build` in this directly.

When I try that, even with the patch, it still fails to compile. But maybe it's because the patch isn't working:

```
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
```

```
> wasm-pack build
[INFO]: 🎯  Checking for the Wasm target...
[INFO]: 🌀  Compiling to Wasm...
warning: patch for `polars` uses the features mechanism. default-features and features will not take effect because the patch dependency does not support this mechanism
warning: Patch `polars v0.22.6 (https://github.com/stuartlynn/polars.git?branch=wasm_rebase#a3632f71)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
   Compiling crossterm v0.23.2
   Compiling zstd-sys v2.0.1+zstd.1.5.2
   Compiling lz4-sys v1.9.3
   Compiling getrandom v0.2.7
   Compiling geozero v0.9.4
error[E0432]: unresolved import `sys::position`
  --> /Users/kyle/.cargo/registry/src/github.com-1ecc6299db9ec823/crossterm-0.23.2/src/cursor.rs:51:9
   |
51 | pub use sys::position;
   |         ^^^^^^^^^^^^^ no `position` in `cursor::sys`

error[E0425]: cannot find value `source` in this scope
  --> /Users/kyle/.cargo/registry/src/github.com-1ecc6299db9ec823/crossterm-0.23.2/src/event/read.rs:24:22
   |
24 |         let source = source.ok().map(|x| Box::new(x) as Box<dyn EventSource>);
   |                      ^^^^^^ a field by this name exists in `Self`

error[E0425]: cannot find function `enable_raw_mode` in module `sys`
   --> /Users/kyle/.cargo/registry/src/github.com-1ecc6299db9ec823/crossterm-0.23.2/src/terminal.rs:120:10
    |
120 |     sys::enable_raw_mode()
    |          ^^^^^^^^^^^^^^^ not found in `sys`

error[E0425]: cannot find function `disable_raw_mode` in module `sys`
   --> /Users/kyle/.cargo/registry/src/github.com-1ecc6299db9ec823/crossterm-0.23.2/src/terminal.rs:127:10
    |
127 |     sys::disable_raw_mode()
    |          ^^^^^^^^^^^^^^^^ not found in `sys`

error[E0425]: cannot find function `size` in module `sys`
   --> /Users/kyle/.cargo/registry/src/github.com-1ecc6299db9ec823/crossterm-0.23.2/src/terminal.rs:134:10
    |
134 |     sys::size()
    |          ^^^^ not found in `sys`

error[E0046]: not all trait items implemented, missing: `eval`
  --> /Users/kyle/.cargo/registry/src/github.com-1ecc6299db9ec823/crossterm-0.23.2/src/event/filter.rs:23:1
   |
6  |     fn eval(&self, event: &InternalEvent) -> bool;
   |     ---------------------------------------------- `eval` from trait
...
23 | impl Filter for EventFilter {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `eval` in implementation

   Compiling rand_core v0.6.3
error[E0308]: mismatched types
   --> /Users/kyle/.cargo/registry/src/github.com-1ecc6299db9ec823/crossterm-0.23.2/src/terminal.rs:104:33
    |
104 | pub fn is_raw_mode_enabled() -> Result<bool> {
    |        -------------------      ^^^^^^^^^^^^ expected enum `std::result::Result`, found `()`
    |        |
    |        implicitly returns `()` as its body has no tail or `return` expression
    |
    = note:   expected enum `std::result::Result<bool, std::io::Error>`
            found unit type `()`

Some errors have detailed explanations: E0046, E0308, E0425, E0432.
For more information about an error, try `rustc --explain E0046`.
error: could not compile `crossterm` due to 7 previous errors
warning: build failed, waiting for other jobs to finish...
The following warnings were emitted during compilation:

warning: liblz4/lib/lz4frame.c:74:10: fatal error: 'stdlib.h' file not found
warning: #include <stdlib.h>   /* malloc, calloc, free */
warning:          ^~~~~~~~~~
warning: liblz4/lib/lz4.c:178:10: fatal error: 'stdlib.h' file not found
warning: #include <stdlib.h>   /* malloc, calloc, free */
warning:          ^~~~~~~~~~
warning: 1 error generated.
warning: In file included from liblz4/lib/lz4hc.c:70:
warning: liblz4/lib/lz4.c:178:10: fatal error: 'stdlib.h' file not found
warning: #include <stdlib.h>   /* malloc, calloc, free */
warning:          ^~~~~~~~~~
warning: 1 error generated.
warning: 1 error generated.
warning: liblz4/lib/xxhash.c:107:10: fatal error: 'stdlib.h' file not found
warning: #include <stdlib.h>
warning:          ^~~~~~~~~~
warning: 1 error generated.

The following warnings were emitted during compilation:

warning: zstd/lib/dictBuilder/cover.c:24:10: fatal error: 'stdio.h' file not found
warning: #include <stdio.h>  /* fprintf */
warning:          ^~~~~~~~~
warning: zstd/lib/dictBuilder/divsufsort.c:39:10: fatal error: 'assert.h' file not found
warning: #include <assert.h>
warning:          ^~~~~~~~~~
warning: 1 error generated.
warning: 1 error generated.
warning: zstd/lib/dictBuilder/fastcover.c:14:10: fatal error: 'stdio.h' file not found
warning: #include <stdio.h>  /* fprintf */
warning:          ^~~~~~~~~
warning: zstd/lib/dictBuilder/zdict.c:41:10: fatal error: 'stdio.h' file not found
warning: #include <stdio.h>         /* fprintf, fopen, ftello64 */
warning:          ^~~~~~~~~
warning: zstd/lib/legacy/zstd_v05.c:1139:10: fatal error: 'stdio.h' file not found
warning: #include <stdio.h>      /* printf (debug) */
warning:          ^~~~~~~~~
warning: 1 error generated.
warning: 1 error generated.
warning: 1 error generated.

error: build failed
Error: Compiling your crate to WebAssembly failed
Caused by: failed to execute `cargo build`: exited with exit status: 101
  full command: "cargo" "build" "--lib" "--release" "--target" "wasm32-unknown-unknown"
```

stuartlynn:
This looks like the best approach for sure. I too was having trouble getting it to compile. It works if I explicitly change the dependency in geopolars (core) to 

```
 polars = { git = "https://github.com/stuartlynn/polars.git", branch = 'wasm_rebase', default-features = false, features = ["ipc", "dtype-u8", "dtype-i8"] }
```
Then the wasm module compiles fine. Patching either in the core Cargo.toml or the wasm Cargo.toml means the build fails. I dont have a lot of experience with the patch feature of Cargo but will see if I can figure out how to resolve this

kylebarron:
Yeah it did look like the patch syntax I used wasn't working

stuartlynn:
Made some progress here. Added some feature flags to geopolars/Cargo.toml which let us retain the features we want from polars when building for regular rust and opt out of them when we are building for wasm.

```rust 
polars = { version = "0.22.6", default-features = false }

[features]
default = [
  "basic-polars",
  "polars/fmt",
  "polars/zip_with",
  "polars/csv-file",
  "polars/temporal",
  "polars/fmt",
  "polars/dtype-slim",
]

basic-polars = ["polars/ipc", "polars/dtype-u8", "polars/dtype-i8", "polars/private"]
```
The basic-polars flag contains the features from polars that we want in both wasm and in regular rust (we prob don't need polars/private but was trying to see if it fixed the new bug). The rest of the features in default are the list of default features for polars.

That means that using the default features for geopolars is equivalent to 
```rust
polars = {version= "0.22.6" features=["ipc","dtype-u8","dtype-i8"]}
```

but when we compile for WASM we can use the patch and just pass in the "basic-polars" feature and skip a lot of the things that give us issues for WASM.

Getting a new error about ArrayRef not being exported from `polars::export::arrow::array` when compiling for WASM but not for regular targets. I think this means either I am missing a feature flag or something about my patch is weird.

Will look in to it more but getting closer here.

wasm-geopolars/Cargo.toml
```diff
@@ -26,7 +26,8 @@ console_error_panic_hook = { version = "0.1.6", optional = true }
 # Unfortunately, `wee_alloc` requires nightly Rust when targeting wasm for now.
 wee_alloc = { version = "0.4.5", optional = true }
 
-geopolars = { path = "../geopolars", default-features = false }
+geopolars = { path = "../geopolars", default-features = false, features = ["basic-polars"] }
```
kylebarron, line 29: Can we add a dependency on polars here so that the polars features needed by wasm are defined here instead of in `geopolars/Cargo.toml`?

stuartlynn:
I think the issue is that Cargo takes a union of the features of dependencies in  each project (https://doc.rust-lang.org/cargo/reference/features.html#feature-unification). So for example if we have 

main Cargo.toml as 
```rust
polars = { version = "0.22.6", features = ["ipc", "dtype-u8", "dtype-i8"] }
```

and 

wasm Cargo.toml as 
```rust
polars = { version = "0.22.6", features = ["ipc", "dtype-u8", "dtype-i8"], default-features=false }
```

it fails because it's taking the union of both of those feature sets which includes things like "fmt" which breaks the wasm build because of it's dependency on crossterm. 

You can see it even clearer if you do 

main Cargo.toml as 
```rust
polars = { version = "0.22.6", features = ["ipc", "dtype-u8", "dtype-i8","fmt"] , default-features=false}
```

and 

wasm Cargo.toml as 
```rust
polars = { version = "0.22.6", features = ["ipc", "dtype-u8", "dtype-i8"], default-features=false }
```

which fails vs 

main Cargo.toml as 
```rust
polars = { version = "0.22.6", features = ["ipc", "dtype-u8", "dtype-i8"] ,default-features=false}
```

and 

wasm Cargo.toml as 
```rust
polars = { version = "0.22.6", features = ["ipc", "dtype-u8", "dtype-i8"], default-features=false }
```

which builds.

Having the feature sets defined in the main Cargo.toml allows is to bypass that because we can specify which features to include  at that level so the union of features for the wasm build never includes them.

There is some discussion of using platform specific dependencies (with resolver=2) here https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies but I am not sure that get's us away from having to define at least part of the solution in the main Cargo.toml.

geopolars/Cargo.toml
```diff
@@ -14,13 +14,30 @@ rust-version = "1.60"
...
 rstar = "0.9.3"
 geo = { git = "https://github.com/georust/geo.git", rev = '578e213875915e1f895892487b5a36ca0d91fac3' }
-polars = { version = "0.22.8", features = ["ipc", "dtype-u8", "dtype-i8"] }
+# Only the _minimal_ features to get polars to compile should be declared here
+polars = { version = "=0.22.6", features = [
+  "dtype-u8",
+  "dtype-i8",
+  "ipc",
+  "private",
+], default-features = false }
```
kylebarron, line 40: I think any features that are _required_ to get the parts of polars we use to compile should be defined here, then more features can be defined above.

stuartlynn (approved):

PR #67, opened 2022-07-16
Improved error handling

kylebarron:
#### Change list

- Create `GeopolarsError` enum in the main crate and `PyGeopolarsError` in the python crate
- Implement `From` so that we can use `?` more broadly
- Implement Python exceptions [as in py-polars](https://github.com/pola-rs/polars/blob/affbeb8f5f920600c40c0ce4c51f89a3c337194d/py-polars/src/error.rs#L56-L62)
- Propagate proj errors in `to_crs`, ref https://github.com/kylebarron/geopolars/pull/62#discussion_r922521616

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -548,6 +548,7 @@ dependencies = [
  "polars",
  "proj",
  "rstar",
+ "thiserror",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


kylebarron:
I'm going to merge to try and keep moving fast, but feedback is still welcome after merge!

PR #73, opened 2022-07-17
Python: Expose rotate, scale, skew

kylebarron:


py-geopolars/src/utils.rs
```diff
@@ -0,0 +1,27 @@
+use crate::error::PyGeopolarsError;
+use geopolars::geoseries::TransformOrigin;
+use pyo3::prelude::*;
+
+#[derive(FromPyObject, Debug)]
+pub(crate) enum PythonTransformOrigin {
+    String(String),  // input is a string
+    Tuple(f64, f64), // input is a 2-tuple of floats
+    Coordinate { x: f64, y: f64 }, // input is a dict of x and y
+}
```
kylebarron, line 10: It looks like `FromPyObject` is a convenient way to handle varied python input: https://pyo3.rs/v0.16.4/conversions/traits.html

PR #80, opened 2022-07-29
Working pyodide build

kylebarron:
For https://github.com/kylebarron/geopolars/issues/77

Notes:

- You have to patch the vendored `chrono` version (in its `cargo.toml`) to 4.19 so it gets picked up correctly. Hopefully this will be unnecessary soon: https://github.com/chronotope/chrono/issues/674
- The geopolars build works and on its own seems to load correctly within pyodide, but it fails at runtime because it tries to subclass `polars` and there isn't a `polars` pyodide wheel on pypi. See related work in https://github.com/pola-rs/polars/issues/3672
- To build `polars`, there might have to be substantial verbose changes in the series bindings because of technical issues (see https://github.com/PyO3/pyo3/issues/2517), there can only be a single `#[pymethods]` implementation. So that might in effect rule out using macros. Such a change probably wouldn't be accepted as a PR into polars.
- Build steps should be the same as described in the polars issue: https://github.com/pola-rs/polars/issues/3672#issuecomment-1188613095


djc:
FYI, chrono 0.4.20 has been released.

kylebarron:
stale and no longer relevant

Issue #48, opened 2022-07-13
Python IO using fiona/OGR


kylebarron:
This will be especially performant once https://github.com/Toblerity/Fiona/pull/1121 is merged, using the latest version of OGR

kylebarron:
For now the read io part has been implemented in #95.

We won't write our own bindings to fiona until the arrow data pr has been merged

jorisvandenbossche:
I was planning to comment here, but I see you also just commented about it at https://github.com/Toblerity/Fiona/pull/1121, so you are aware of it.

So you can already get the ArrayStream pointer using "just" the GDAL python API, without the need for fiona or pyogrio (https://github.com/geopandas/pyogrio/pull/155). 
That might be something you could also directly use here? 

Of course to use that, you need to have those Python bindings installed. What fiona or pyogrio give you is a way to install GDAL, since they include GDAL in their wheels, but if you get GDAL that way, you don't have the python bindings (and for this installation use case, you indeed need an API in fiona or pyogrio). But for example if you are using conda and have installed GDAL that way, the python bindings are also installed and can be used.

kylebarron:
Looking at https://github.com/geopandas/pyogrio/pull/155 also makes me think it's wise to let people more knowledgeable on C and GDAL work through the gritty parts 😅

PR #95, opened 2022-08-28
read io wrapping geopandas

kylebarron:
This adds:

- `geopolars.read_file`
- `geopolars.from_geopandas` (not yet implemented for GeoSeries)
- `GeoDataFrame._from_geopandas`

```py
import geopandas as gpd

import geopolars as gpl

path = gpd.datasets.get_path("nybb")
gdf = gpl.read_file(path)
```


kylebarron:
For https://github.com/geopolars/geopolars/issues/48

py-geopolars/python/geopolars/io/file.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def read_file(*args, **kwargs):
+    if geopandas is None:
+        raise ImportError(
+            "Geopandas is currently required for the read_file method. Install it with `pip install geopandas`."
+        )
+
+    geopandas_gdf = geopandas.read_file(*args, **kwargs)
+    arrow_table = geopandas.io.arrow._geopandas_to_arrow(geopandas_gdf)
```
brendan-ward, line 18: This makes me think we should promote this function to part of the public GeoPandas API as something like a `to_arrow` method.  @ jorisvandenbossche?

kylebarron: A public API to import/export geopandas to in-memory arrow data would be useful.

PR #99, opened 2022-09-07
Fix python example, return GeoSeries from get_column

mchalecki:
This PR adresses two things
- Fix example.py as this wasn't working. Now it returns correct centroid of dataframe. Previously it failed on import because centroid is a method on GeoSeries.
- GeoSeries.get_column("geometry") was returning Series instead of GeoSeries. This pr modify get_column to return GeoSeries when requested column is geo column

### DIFF
--- a/py-geopolars/python/geopolars/internals/geodataframe.py
+++ b/py-geopolars/python/geopolars/internals/geodataframe.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from polars import DataFrame
+import polars as pl
 
 from geopolars import geopolars as core  # type: ignore
 from geopolars.internals.geoseries import GeoSeries
@@ -8,18 +8,45 @@ from geopolars.internals.geoseries import GeoSeries
 DEFAULT_GEO_COLUMN_NAME = "geometry"
 
 
-class GeoDataFrame(DataFrame):
+class GeoDataFrame(pl.DataFrame):
 
     _geometry_column_name = DEFAULT_GEO_COLUMN_NAME
 
     def __init__(self, data=None, columns=None, orient=None, *, geometry=None):
 
         # Wrap an existing polars DataFrame
-        if isinstance(data, DataFrame):
+        if isinstance(data, pl.DataFrame):
             self._df = data._df
             return
 
         super().__init__(data, columns, orient)
+
+    def get_column(self, name: str) -> pl.Series | GeoSeries:
+        """
+        Get a single column as Series or GeoSeries by name. Return GeoSeries if requested column is geometry column.
+
+        Parameters
+        ----------
+        name : str
+            Name of the column to retrieve.
+
+        Examples
+        --------
+        >>> df = pl.DataFrame({"foo": [1, 2, 3], "bar": [4, 5, 6]})
+        >>> df.get_column("foo")
+        shape: (3,)
+        Series: 'foo' [i64]
+        [
+                1
+                2
+                3
+        ]
+
+        """
+        series = super().get_column(name)
+        if name == self._geometry_column_name:
+            series = GeoSeries(series)
+        return series
 
     @property
     def geometry(self):
--- a/py-geopolars/python/geopolars/internals/geodataframe.py
+++ b/py-geopolars/python/geopolars/internals/geodataframe.py
@@ -1,18 +1,21 @@
-import polars
+from __future__ import annotations
 
+from polars import DataFrame
+
+from geopolars import geopolars as core  # type: ignore
 from geopolars.internals.geoseries import GeoSeries
 
 DEFAULT_GEO_COLUMN_NAME = "geometry"
 
 
-class GeoDataFrame(polars.DataFrame):
+class GeoDataFrame(DataFrame):
 
     _geometry_column_name = DEFAULT_GEO_COLUMN_NAME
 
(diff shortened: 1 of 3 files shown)


mchalecki:
I don't like multiple types of return - that the get_column might return GeoSeries or Series, but it is what geopandas does, so I followed.
```python
In [6]: x = gpd.GeoDataFrame({"z":[1,2,3,4], "y":[2,3,4,5]})

In [7]: type(x["z"])
Out[7]: pandas.core.series.Series
```

kylebarron:
Thanks for the PR! Yes, the `example.py` file was outdated. It was one of the first Python files, added way back in https://github.com/geopolars/geopolars/pull/3/files#diff-53512b24f7fdf18d6efd312b230dbdb2d695fcac421c825226e3e76f69094b39 to test that the Python-Rust binding was working at all. At that time, the exported Python API was just the pure Rust functions, e.g. what you see in https://github.com/geopolars/geopolars/blob/16c3afe0954373bea9746b70a0df6b9dfe3c8f98/py-geopolars/src/geoseries.rs. (So for example `example.py` probably would've worked if you changed the import to `from geopolars.geopolars import centroid`, but now that's a private API (though I haven't figured out how to change that underlying module name to `geopolars._geopolars`). Now we've added a Python part so that we can subclass `polars.Series` and that file hadn't gotten updated yet.

py-geopolars/python/geopolars/internals/geodataframe.py
```diff
@@ -1,23 +1,53 @@
...
 
         # Wrap an existing polars DataFrame
-        if isinstance(data, polars.DataFrame):
+        if isinstance(data, pl.DataFrame):
             self._df = data._df
             return
 
         super().__init__(data, columns, orient)
 
+    def get_column(self, name: str) -> pl.Series | GeoSeries:
```
kylebarron, line 24: I've always used `Union[...]` instead of `A | B`. Can you remind me what's the first Python version that supports `A | B`? I had thought it was higher than 3.7, which we currently support.

py-geopolars/src/api.rs
```diff
@@ -1,4 +1,4 @@
-use crate::geoseries;
+use crate::geefoseries;
```
kylebarron, line 1: typo?

py-geopolars/python/geopolars/internals/geodataframe.py
```diff
@@ -1,23 +1,53 @@
...
+        Series: 'foo' [i64]
+        [
+                1
+                2
+                3
+        ]
+
+        """
+        series = super().get_column(name)
+        if name == self._geometry_column_name:
```
kylebarron, line 47: 👍 In the future, when polars supports extension types, we should be able to handle multiple geometry columns (and not forcing `_geometry_column_name`) by checking if the returned series is an extension type of type `geoarrow`

kylebarron (approved):

py-geopolars/python/geopolars/internals/geodataframe.py
kylebarron: Hmm so it looks like you _can_ use it in Python 3.7 annotations, but you can never let Python evaluate the types (it will only be readable by mypy). https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-x-y-syntax-for-unions

polars seems to use `|` extensively, and in general I'd try to align with their style, so I'd say this is ok 👍

py-geopolars/python/geopolars/internals/geodataframe.py
mchalecki: sure, or have property returning names of geo columns. But for now i think it is enough to check this condition

py-geopolars/python/geopolars/internals/geodataframe.py
```diff
@@ -1,23 +1,53 @@
...
+import polars as pl
+
+from geopolars import geopolars as core  # type: ignore
 from geopolars.internals.geoseries import GeoSeries
 
 DEFAULT_GEO_COLUMN_NAME = "geometry"
 
 
-class GeoDataFrame(polars.DataFrame):
+class GeoDataFrame(pl.DataFrame):
```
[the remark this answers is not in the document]
mchalecki: Sure. Tell me if I should change it. I just assumed that because for pandas we use `pd` in `example.py` is used `import geopolars as gpl`. If there are some set conventions i would gladly use it.

py-geopolars/src/api.rs
mchalecki: Yes, typo sorry

mchalecki:
I just started to get to know this project and though binding are good place to start as I have more practice with python than rust. Do you see any particular issue that would be nice as begging ?

kylebarron:
There's a lot of stuff to be done on the Python side. I think the first is to set up the structure for python tests https://github.com/geopolars/geopolars/issues/100. Then after that I've started work on interop between geopolars and geopandas https://github.com/geopolars/geopolars/pull/95. Also I've started on python sphinx docs https://github.com/geopolars/geopolars/tree/master/py-geopolars/docs, borrowing from both geopandas and polars docs, but that still needs to get cleaned up a bit and then get published to github pages.

There's a lot to do so whatever you're most interested in is fair game. Open/comment on an issue and let's discuss

PR #109, opened 2022-09-15
chore(python): build arm64 wheels instead of universal2

kylebarron:
Originally I thought there was a bug in maturin that incorrectly used the x86 name when building arm64 wheels, but it appears that bug is gone now. Building x86 and arm wheels separately means that more files are produced, but each one is half the size.

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -3,8 +3,8 @@ name: Build Python Wheels
 # Only run on new tags starting with `v`
 on:
   push:
-    # tags:
-    #   - "v*"
+    tags:
+      - "v*"
 
 jobs:
   build_wheels:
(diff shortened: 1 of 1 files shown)


kylebarron:
>       thread 'main' panicked at 'Unable to find libclang: "couldn't find any valid shared libraries matching: ['libclang.so', 'libclang-*.so', 'libclang.so.*', 'libclang-*.so.*'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [])"', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/bindgen-0.59.2/src/lib.rs:2144:31

Might need to add a clang install step to ci

kylebarron:
it turns out the clang error is just from PROJ. Disabling that in https://github.com/geopolars/geopolars/pull/126 to be able to make releases

PR #110, opened 2022-09-18
Implementation of distance on GeoSeries

mchalecki:
- Implementation of 1:1 distance on GeoSeries
- Fix deprecated in ffi acquire_gil
- Always return python GeoSeries when using getitem
- Fix typing in test fixture

### DIFF
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -653,9 +653,7 @@ impl GeoSeries for Series {
         let mut output_array = MutablePrimitiveArray::<f64>::with_capacity(self.len());
 
         for (g1, g2) in iter_geom(self).zip(iter_geom(other)) {
-            let d2=g1.euclidean_distance(g2);
             let distance = match (g1, g2) {
-                (Geometry::Point(p1), p2) => Some(p1.euclidean_distance(&p2)),
                 (Geometry::Point(p1), Geometry::Point(p2)) => Some(p1.euclidean_distance(&p2)),
                 (Geometry::Point(p1), Geometry::LineString(p2)) => Some(p1.euclidean_distance(&p2)),
                 (Geometry::Point(p1), Geometry::MultiLineString(p2)) => {
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -162,6 +162,13 @@ pub trait GeoSeries {
     /// ```
     fn skew(&self, xs: f64, ys: f64, origin: TransformOrigin) -> Result<Series>;
 
+    /// Returns a Series containing the distance to aligned other. The operation works on a 1-to-1 row-wise manner.
+    ///
+    /// # Arguments
+    ///
+    /// * `other` - The Geoseries (elementwise) or geometric object to find the distance to.
+    fn distance(&self, other: &Series) -> Result<Series>;
+
     // Note: Ideally we wouldn't have both `from` and `to` here, where the series would include the
     // current CRS, but that would require polars to support extension types.
     #[cfg(feature = "proj")]
@@ -640,6 +647,35 @@ impl GeoSeries for Series {
         }
     }
 
+    fn distance(&self, other: &Series) -> Result<Series> {
+        use geo::algorithm::EuclideanDistance;
+
+        let mut output_array = MutablePrimitiveArray::<f64>::with_capacity(self.len());
+
+        for (g1, g2) in iter_geom(self).zip(iter_geom(other)) {
+            let distance = match (g1, g2) {
+                (Geometry::Point(p1), Geometry::Point(p2)) => Some(p1.euclidean_distance(&p2)),
+                (Geometry::Point(p1), Geometry::LineString(p2))=>Some(p1.euclidean_distance(&p2)),
+                (Geometry::Point(p1), Geometry::MultiLineString(p2))=>Some(p1.euclidean_distance(&p2)),
+                (Geometry::Point(p1), Geometry::MultiPoint(p2))=>Some(p1.euclidean_distance(&p2)),
+                (Geometry::Point(p1), Geometry::MultiPolygon(p2))=>Some(p1.euclidean_distance(&p2)),
+                (Geometry::Point(p1), Geometry::Polygon(p2))=>Some(p1.euclidean_distance(&p2)),
+
+                (Geometry::LineString(p1), Geometry::Point(p2))=>Some(p2.euclidean_distance(&p1)),
+                (Geometry::MultiLineString(p1), Geometry::Point(p2))=>Some(p2.euclidean_distance(&p1)),
+                (Geometry::MultiPoint(p1), Geometry::Point(p2))=>Some(p2.euclidean_distance(&p1)),
+                (Geometry::MultiPolygon(p1), Geometry::Point(p2))=>Some(p2.euclidean_distance(&p1)),
+                (Geometry::Polygon(p1), Geometry::Point(p2))=>Some(p2.euclidean_distance(&p1)),
+                _ => None,
+            };
+            output_array.push(distance);
+        }
+
+        let result: PrimitiveArray<f64> = output_array.into();
+        let series = Series::try_from(("distance", Box::new(result) as ArrayRef))?;
+        Ok(series)
+    }
+
     #[cfg(feature = "proj")]
     fn to_crs(&self, from: &str, to: &str) -> Result<Series> {
         use proj::{Proj, Transform};
@@ -718,9 +754,11 @@ mod tests {
     use polars::prelude::Series;
     use std::sync::Arc;
 
-    use geo::{line_string, polygon, CoordsIter, Geometry, LineString, MultiPoint, Point};
+    use geo::{line_string, polygon, CoordsIter, Geometry, LineString, MultiPoint, Point, point, Line};
     use geozero::{CoordDimensions, ToWkb};
-    use polars::export::arrow::array::{ArrayRef, BinaryArray, MutableBinaryArray};
+    use polars::export::arrow::array::{
+        ArrayRef, BinaryArray, MutableBinaryArray, MutablePrimitiveArray, PrimitiveArray,
+    };
 
     use super::TransformOrigin;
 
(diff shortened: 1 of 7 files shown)


mchalecki:
I had all precommit passed
```
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
black....................................................................Passed
isort....................................................................Passed
pyupgrade................................................................Passed
flake8...................................................................Passed
```

kylebarron:
> I had all precommit passed

Yeah `pre-commit` only runs the python lints at the moment. The rust lints are manual through `cargo fmt --all` (formatting) and `cargo clippy --all --all-features --tests` (lints)

py-geopolars/python/geopolars/internals/geoseries.py
```diff
@@ -26,6 +29,19 @@ def __init__(self, *args, **kwargs):
...
+        | pl.Series
+        | GeoSeries
+        | range
+        | slice
+        | np.ndarray[Any, Any]
+        | list[int]
+        | list[bool],
+    ) -> Any:
+        return GeoSeries(super().__getitem__(item))
+
```
kylebarron, line 44: This means that if you do `GeoSeries(data)[0]` you get a `GeoSeries` back? I think of a `GeoSeries` as representing a collection of geometries, so having it also represent a single scalar seems confusing.

It also seems to be inconsistent with upstream polars where `__getitem__` returns a scalar value.

```py
import polars as pl
s = pl.Series([1, 2, 3, 4])
type(s[0]) # int
```

It's possible we'll need to handle single geometries with a `Geometry` class, but I think that would need more thought for how to implement that in a way that doesn't end up storing separate memory objects for each geometry.

py-geopolars/src/geoseries.rs
```diff
@@ -141,6 +141,19 @@ pub(crate) fn skew(
...
+pub(crate) fn distance(
+    series: &PyAny,
+    other: &PyAny,
+) -> PyResult<PyObject> {
+    let series = ffi::py_series_to_rust_series(series)?;
+    let other = ffi::py_series_to_rust_series(other)?;
+    let out = series
+        .distance(&other)
+        .map_err(PyGeopolarsError::from)?;
+    ffi::rust_series_to_py_geoseries(&out)
```
kylebarron, line 154: This should be `rust_series_to_py_series` I believe, as the return values is non-geographic

geopolars/src/geoseries.rs
```diff
@@ -162,6 +162,13 @@ pub trait GeoSeries {
     /// ```
     fn skew(&self, xs: f64, ys: f64, origin: TransformOrigin) -> Result<Series>;
 
+    /// Returns a Series containing the distance to aligned other. The operation works on a 1-to-1 row-wise manner.
+    ///
+    /// # Arguments
+    ///
+    /// * `other` - The Geoseries (elementwise) or geometric object to find the distance to.
```
kylebarron, line 169: > or geometric object 

Does this work with a single geometry passed in as `other`? It looks like it needs other to be another GeoSeries

geopolars/src/geoseries.rs
```diff
@@ -162,6 +162,13 @@ pub trait GeoSeries {
     /// ```
     fn skew(&self, xs: f64, ys: f64, origin: TransformOrigin) -> Result<Series>;
 
+    /// Returns a Series containing the distance to aligned other. The operation works on a 1-to-1 row-wise manner.
```
kylebarron, line 165: It would be great for the docstring for here and for the Python API to make clear that it's calculating cartesian distance in 2D space, and the units of the output are in terms of the CRS of the two input series. (I'm surprised the geopandas docstring doesn't mention that)

geopolars/src/geoseries.rs
```diff
@@ -640,6 +647,35 @@ impl GeoSeries for Series {
...
+                (Geometry::Point(p1), Geometry::MultiPoint(p2))=>Some(p1.euclidean_distance(&p2)),
+                (Geometry::Point(p1), Geometry::MultiPolygon(p2))=>Some(p1.euclidean_distance(&p2)),
+                (Geometry::Point(p1), Geometry::Polygon(p2))=>Some(p1.euclidean_distance(&p2)),
+
+                (Geometry::LineString(p1), Geometry::Point(p2))=>Some(p2.euclidean_distance(&p1)),
+                (Geometry::MultiLineString(p1), Geometry::Point(p2))=>Some(p2.euclidean_distance(&p1)),
+                (Geometry::MultiPoint(p1), Geometry::Point(p2))=>Some(p2.euclidean_distance(&p1)),
+                (Geometry::MultiPolygon(p1), Geometry::Point(p2))=>Some(p2.euclidean_distance(&p1)),
+                (Geometry::Polygon(p1), Geometry::Point(p2))=>Some(p2.euclidean_distance(&p1)),
+                _ => None,
```
kylebarron, line 669: What happens when neither of the two geometries are of type `Point`? There are `impl`s for it looks like every combination of geometries: https://docs.rs/geo/latest/geo/algorithm/euclidean_distance/trait.EuclideanDistance.html#impl-EuclideanDistance%3CT%2C%20LineString%3CT%3E%3E-for-Polygon%3CT%3E

I don't know the best way to handle these functions where we want to support the cartesian product between each geometry type and each other geometry type. I hope we don't have to manually list out every combination of geometry types every time, like we do here:
https://github.com/geopolars/geopolars/blob/6c28420581866fb38745dd6044c615a3f4b93c2c/geopolars/src/spatial_index.rs#L86-L132

geopolars/src/geoseries.rs
```diff
@@ -640,6 +647,35 @@ impl GeoSeries for Series {
         }
     }
 
+    fn distance(&self, other: &Series) -> Result<Series> {
+        use geo::algorithm::EuclideanDistance;
+
+        let mut output_array = MutablePrimitiveArray::<f64>::with_capacity(self.len());
+
+        for (g1, g2) in iter_geom(self).zip(iter_geom(other)) {
```
kylebarron, line 655: Curious whether we should have an assertion before this that both series have the same length. Thoughts?

kylebarron (commented):
In https://github.com/geopolars/geopolars/pull/42 @stuartlynn put up a proposal for handling both pairwise Series-Series and broadcasted Series-Scalar function handling. In the future we'll want `distance` to support `other` being a scalar as well, but I don't think that's a blocker; it's fine to have series-series support only

py-geopolars/python/geopolars/internals/geoseries.py
mchalecki: Yea. It's a miss in your case when returned a scalar. There was the case when I selected range from geometry column and got back slice of geometry columns but as a series now

What do you thing about returning Geoseries if we were to return Series?

mchalecki:
> In #42 @stuartlynn put up a proposal for handling both pairwise Series-Series and broadcasted Series-Scalar function handling. In the future we'll want `distance` to support `other` being a scalar as well, but I don't think that's a blocker; it's fine to have series-series support only

I agree. It didn't have an idea how to implement second behavior so i started with this implemented one in geo, the one that geopandas also uses

py-geopolars/python/geopolars/internals/geoseries.py
mchalecki: Btw what is initial plan to represent geometries in python?
Right now in most cases they are just list of u8. In geopandas shapely is used to represent geom classes like Point and others.

kylebarron: Ah I see what you mean. For _nested_ types, the `__getitem__` of a series is another series

```py
import geopolars as gpl
geom_col = gpl.datasets.read_dataset('naturalearth_cities').get_column('geometry')

type(geom_col[0])
# polars.internals.series.series.Series

geom_col[0].dtype
# polars.datatypes.UInt8
```

The issue is that this sub-series no longer fits in with `GeoSeries`; all of the `GeoSeries` methods would fail because `GeoSeries` expects that each element is an entire geometry, not a `u8`.

> Btw what is initial plan to represent geometries in python?

This is hard to say. I think we really want to stick with Arrow memory for geometries, which means that we can't expose geometries directly as Shapely objects. Maybe it would make sense to add geometry class(es) that wrap individual objects, but we'd have to be careful to ensure the memory is a view on existing column memory

As further background: for now in Rust we use store geometries as WKB (with a `List<u8>` array type because Polars doesn't support the `Binary` type directly) but that's only a stopgap measure. We really want to use the [geo-arrow native encoding](https://github.com/geopandas/geo-arrow-spec/blob/main/format.md), so as soon as Polars supports `FixedSizeList` we'll switch over to that, which should make interop with Shapely geometries faster. E.g. Joris benchmarked to/from arrow to be significantly faster than to/from WKB. https://github.com/jorisvandenbossche/python-geoarrow

geopolars/src/geoseries.rs
mchalecki: i will add this feature

kylebarron: Before supporting both, I think it would be good to first discuss and merge https://github.com/geopolars/geopolars/pull/42, so that we have a broadcasting approach that will work with all functions

geopolars/src/geoseries.rs
mchalecki: Good idea. Not sure how to pass gracefully exception to python.

geopolars/src/geoseries.rs
mchalecki: Good idea cause i was optimistic at first. It is easy to calculate against single value. But after a thought i realized that I would need to have 3 functions instead of one i think as there is no easy way for overload. So broadcasting would be nice.

Ok i will leave it as it is

py-geopolars/python/geopolars/internals/geoseries.py
mchalecki: Thanks for clarification. Is there other place for discussion e.g slack, discrod (as polars, or on polars server) ?

So perhaps I'll remove changes, leave as it was and create a separate issue

geopolars/src/geoseries.rs
mchalecki: for some reason i though that I had them all, didnt see impl for other classes.

I don't know the better way. I really wanted to do it differently because for sure it doesn't scale. More so that It is symetric relation so if there is no implementation X,Y=d i need to check if there is Y,X=d. And i think it happens more often.

And all of those types use the same stuff e.g `poly_lhs.intersects(poly_rhs)`. I dont know how in rust you would work with that. I guess some macro would be nice

mchalecki:
Can I add in black line length 120 instead of default 80?

kylebarron:
> Can I add in black line length 120 instead of default 80?

I'd prefer to stay as close to the upstream py-polars settings as possible, and they use the default of 88 I believe.

py-geopolars/python/geopolars/internals/geoseries.py
kylebarron: > Thanks for clarification. Is there other place for discussion e.g slack, discrod (as polars, or on polars server) ?

So far we haven't created another place for discussions, so I think issues or [discussions](https://github.com/geopolars/geopolars/discussions) on this repo is the easiest solution

> So perhaps I'll remove changes, leave as it was and create a separate issue

Yeah a new issue would be great for this!

geopolars/src/geoseries.rs
mchalecki: @kylebarron  I can create such macro in another PR if you specify it how it should work because that two cases differ a bit. For now i added missing traits. But i dont like it that had to copy paste now the same stuff.

kylebarron: I don't have much experience with Rust macros. Definitely not necessary for this PR. Eventually I'd like to figure out a better solution

geopolars/src/geoseries.rs
```diff
@@ -707,6 +759,27 @@ impl GeoSeries for Series {
         let series = Series::try_from(("result", Box::new(result) as ArrayRef))?;
         Ok(series)
     }
+
+    #[cfg(feature = "proj")]
+    fn to_crs(&self, from: &str, to: &str) -> Result<Series> {
```
kylebarron, line 764: Can you move `to_crs` back to where it was for a smaller diff?

kylebarron (approved):

geopolars/src/geoseries.rs
mchalecki: Asserts are used currently only in tests. I can add it here but i guess it will fail spectacularly in python

geopolars/src/geoseries.rs
mchalecki: sorry. Fixed

kylebarron (approved):

stuartlynn:
> > In #42 @stuartlynn put up a proposal for handling both pairwise Series-Series and broadcasted Series-Scalar function handling. In the future we'll want `distance` to support `other` being a scalar as well, but I don't think that's a blocker; it's fine to have series-series support only
> 
> I agree. It didn't have an idea how to implement second behavior so i started with this implemented one in geo, the one that geopandas also uses

Hey @mchalecki ! Thanks for this, hoping to finish up that proposal this week and then implement it for each of the functions we already have. Will add this one to the list!

Issue #31, opened 2022-06-09
Geometry representation and zero-copy access


kylebarron:
At this point, to get things started, Well Known Binary geometries are stored in an Arrow binary array. But this is an inefficient in-memory format. For one, WKB is not zero copy; offsets (such as where rings start) are not known until the buffer is parsed.

The ideal memory format would be an [arrow-native geometry encoding](https://github.com/geopandas/geo-arrow-spec/blob/main/format.md). With Arrow-native arrays, any coordinate of any geometry in the column can be accessed with zero copy and in constant time.

But the `geo` crate revolves around its own geometry structs, defined in `geo-types`... So today, whether we store geometries in an array of WKB or in an arrow-native encoding, either way they need to be parsed and copied into `geo-types` structs.

I've been talking to the `georust` project a bit (on discord) about reviving their consideration of _geometry traits_. That is, instead of the `georust` algorithms requiring _`geo-types` structs_ as input, it would instead require a trait that implements something like a `Point`, etc.

This would be super powerful here because we could implement access traits for geometries stored in arrow buffers. The best issue to follow is probably https://github.com/georust/geo/discussions/838

kylebarron:
cc @stuartlynn

kylebarron:
> It sounds like the geo-types as traits is further off than the arrow-native geometry encoding

I think a `geo-types` trait is indeed a bit of a long-run goal here. But it's not clear to me that an arrow-native geometry encoding is that worthwhile _without_ having the `geo-types` traits. Because then you'll still need to copy the arrow memory into a `geo` object for every row on every algorithm.

A stopgap measure could also be to make use of the polars [`ObjectArray`](https://docs.rs/polars/latest/polars/chunked_array/object/struct.ObjectArray.html) which can wrap [any Rust object](https://docs.rs/polars/latest/polars/chunked_array/object/trait.PolarsObject.html). But this is just a stopgap because it doesn't store the geometries in Arrow memory, so it isn't able to be shared with e.g. pyarrow.

> Not sure if there is a reference implementation being worked on just now for the arrow native encoding

There are a couple implementations in progress e.g. https://github.com/jorisvandenbossche/python-geoarrow, https://github.com/paleolimbot/geoarrow, https://github.com/paleolimbot/geoarrow-cpp.

> We would also presumably need to figure out how this interacts with the discussion of custom extension types in polars

[...]

PR #111, opened 2022-09-18
Implement `explode_geometries` method

### LINKED ISSUE #26: Implement: Explode
This should take a GeoSeries with multipart geometries (MultiPoint, MultiPolygon etc) and return a GeoSeries with each individual geometry in the multi collection as an individual geometry

Is this implemented in the Geo crate: No
GeoPandas equivalent: [explode](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.explode.html)

Note this could be done by simply iterating through the series and then each Multi Geometry and appending the geometries to a new series. 

One thing to figure out is how we want to handle this in the context of a GeoDataFrame. We would want the resulting geometries to have the same properties as their parent geometries.

In Geopandas, the result is a Series with a mutli-level index, we might not need this for polars

edpft:
Closes #26 by implementing an `explode_geometries` method that takes a `GeoSeries` with multipart geometries and return a `GeoSeries` with those geometries split into multiple individual geometries.

I've taken the approach suggested in the original issue, of iterating through the geometries and appending them to a new series.

I've added a test cribbed from the `geopandas` example to prove this works and a very minimal docstring, also copied from `geopandas`.

A few things to note:

- I went with `explode_geometries` instead of `explode` because when I used `explode` I was just getting the `polars` method of the same name. This is probably just ignorance on my part!
- I'm not handling  `Geometry::GeometryCollection` because I wasn't sure if I should be but I could do this recursively, if I abstract the match expression into a function.
- My naming conversions are a little more verbose than yours (e.g. `geometry` instead of `geom`) but I'd be happy to conform if needed

For what it's worth, If I was going to implement this for your `GeoDataFrame`, I'd probably duplicate the other information and add a "geometry index" `Series` with the index of the geometry within the original multipart geometry.

### DIFF
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -71,7 +71,7 @@ pub trait GeoSeries {
     fn exterior(&self) -> Result<Series>;
 
     /// Explodes multi-part geometries into multiple single geometries.
-    fn explode_geometries(&self) -> Result<Series>;
+    fn explode(&self) -> Result<Series>;
 
     /// Create a Series from a vector of geometries
     fn from_geom_vec(geoms: &[Geometry<f64>]) -> Result<Series>;
@@ -310,7 +310,7 @@ impl GeoSeries for Series {
         Ok(series)
     }
 
-    fn explode_geometries(&self) -> Result<Series> {
+    fn explode(&self) -> Result<Series> {
         let mut exploded_vector = Vec::new();
 
         for geometry in iter_geom(self) {
@@ -1002,7 +1002,7 @@ mod tests {
         assert_eq!(10.0_f64, as_vec[0]);
     }
     #[test]
-    fn explode_geometries() {
+    fn explode() {
         let point_0 = Point::new(0., 0.);
         let point_1 = Point::new(1., 1.);
         let point_2 = Point::new(2., 2.);
@@ -1027,7 +1027,7 @@ mod tests {
         ])
         .unwrap();
 
-        let output_series = input_series.explode_geometries().unwrap();
+        let output_series = GeoSeries::explode(&input_series).unwrap();
 
         assert_eq!(output_series, expected_series);
     }
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -311,7 +311,7 @@ impl GeoSeries for Series {
     }
 
     fn explode_geometries(&self) -> Result<Series> {
-        let mut exploded_vector = vec![];
+        let mut exploded_vector = Vec::new();
 
         for geometry in iter_geom(self) {
             match geometry {
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -311,55 +311,57 @@ impl GeoSeries for Series {
     }
 
     fn explode_geometries(&self) -> Result<Series> {
-        let mut nested_vector = vec![];
+        let mut exploded_vector = vec![];
 
         for geometry in iter_geom(self) {
             match geometry {
                 Geometry::Point(geometry) => {
-                    let point = vec![Geometry::Point(geometry)];
-                    nested_vector.push(point)
+                    let point = Geometry::Point(geometry);
+                    exploded_vector.push(point)
                 }
                 Geometry::MultiPoint(geometry) => {
-                    let points: Vec<Geometry> = geometry.into_iter().map(Geometry::Point).collect();
-                    nested_vector.push(points)
+                    for geom in geometry.into_iter() {
+                        let point = Geometry::Point(geom);
+                        exploded_vector.push(point)
+                    }
                 }
                 Geometry::Line(geometry) => {
-                    let line = vec![Geometry::Line(geometry)];
-                    nested_vector.push(line)
+                    let line = Geometry::Line(geometry);
+                    exploded_vector.push(line)
                 }
                 Geometry::LineString(geometry) => {
-                    let line_string = vec![Geometry::LineString(geometry)];
-                    nested_vector.push(line_string)
+                    let line_string = Geometry::LineString(geometry);
+                    exploded_vector.push(line_string)
                 }
                 Geometry::MultiLineString(geometry) => {
-                    let line_strings: Vec<Geometry> =
-                        geometry.into_iter().map(Geometry::LineString).collect();
-                    nested_vector.push(line_strings)
+                    for geom in geometry.into_iter() {
+                        let line_string = Geometry::LineString(geom);
+                        exploded_vector.push(line_string)
+                    }
                 }
                 Geometry::Polygon(geometry) => {
-                    let polygon = vec![Geometry::Polygon(geometry)];
-                    nested_vector.push(polygon)
+                    let polygon = Geometry::Polygon(geometry);
+                    exploded_vector.push(polygon)
                 }
                 Geometry::MultiPolygon(geometry) => {
-                    let polygons: Vec<Geometry> =
-                        geometry.into_iter().map(Geometry::Polygon).collect();
-                    nested_vector.push(polygons)
+                    for geom in geometry.into_iter() {
+                        let polygon = Geometry::Polygon(geom);
+                        exploded_vector.push(polygon)
+                    }
                 }
                 Geometry::Rect(geometry) => {
-                    let rectangle = vec![Geometry::Rect(geometry)];
-                    nested_vector.push(rectangle)
+                    let rectangle = Geometry::Rect(geometry);
+                    exploded_vector.push(rectangle)
                 }
                 Geometry::Triangle(geometry) => {
-                    let triangle = vec![Geometry::Triangle(geometry)];
-                    nested_vector.push(triangle)
+                    let triangle = Geometry::Triangle(geometry);
+                    exploded_vector.push(triangle)
                 }
                 _ => unimplemented!(),
             };
         }
 
-        let flattened_vector: Vec<Geometry> = nested_vector.into_iter().flatten().collect();
-
-        let exploded_series = Series::from_geom_vec(&flattened_vector)?;
+        let exploded_series = Series::from_geom_vec(&exploded_vector)?;
 
         Ok(exploded_series)
     }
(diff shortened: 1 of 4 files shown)


geopolars/src/geoseries.rs
```diff
@@ -70,6 +70,9 @@ pub trait GeoSeries {
     /// Applies to GeoSeries containing only Polygons. Returns `None` for other geometry types.
     fn exterior(&self) -> Result<Series>;
 
+    /// Explodes multi-part geometries into multiple single geometries.
+    fn explode_geometries(&self) -> Result<Series>;
```
kylebarron, line 74: I see that polars [already has an `explode` function](https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.Series.explode.html?highlight=explode#polars.Series.explode), but in this case, the meaning is the same for both, so maybe we could call this fn `explode` as well?

geopolars/src/geoseries.rs
```diff
@@ -307,6 +310,60 @@ impl GeoSeries for Series {
...
+
+        for geometry in iter_geom(self) {
+            match geometry {
+                Geometry::Point(geometry) => {
+                    let point = vec![Geometry::Point(geometry)];
+                    nested_vector.push(point)
+                }
+                Geometry::MultiPoint(geometry) => {
+                    let points: Vec<Geometry> = geometry.into_iter().map(Geometry::Point).collect();
+                    nested_vector.push(points)
```
kylebarron, line 324: Instead of having `nested_vector` being `vec![vec![geometry]]`, could we push individual parts into that function?
```rs
for geom in geometry.into_iter() {
    output_vector.push(Geometry::Point(geom))
}
```

That means
1. we avoid the extra vec construction for each geometry
2. we avoid an extra iteration down below when you flatten

kylebarron:
Thanks for the PR!

I've never actually used `explode` from GeoPandas, so I'm just reading up on it now. My first thought is that it might make sense for us to consider the `GeoDataFrame.explode` and `GeoSeries.explode` implementations at the same time, so that we don't implement it for `Series` and then find that it's difficult to implement for `GeoDataFrame`.

In GeoPandas, it looks like `explode` sets a multiindex to keep track of which rows the geometries came from. Polars doesn't have a concept of an index, so we'll have to record indexes separately somehow.

> * I went with `explode_geometries` instead of `explode` because when I used `explode` I was just getting the `polars` method of the same name. This is probably just ignorance on my part!

Hmm I have to look up the syntax but I _think_ there's a way for users to declare which trait a function comes from. So if we name it the same, I think users would have to "opt-in" to the `GeoSeries` trait's `explode`

> * I'm not handling  `Geometry::GeometryCollection` because I wasn't sure if I should be but I could do this recursively, if I abstract the match expression into a function.

That's ok. I think for now it's ok to ignore GeometryCollection and we can come back in the future when we want to make things more stable.

> For what it's worth, If I was going to implement this for your `GeoDataFrame`, I'd probably duplicate the other information and add a "geometry index" `Series` with the index of the geometry within the original multipart geometry.

For the `GeoDataFrame` case, I'm curious if there's an existing polars function that would allow us to provide a list of indexes and have polars automatically duplicate rows. Maybe we could leverage join here somehow

brendan-ward:
I'm not yet familiar enough with GeoPolars to be sure I'm saying the _right_ thing here, so take this with a grain of salt - these are just thoughts from the GeoPandas implementation in case those are useful here (I regularly use `.explode()` and the underlying `get_parts`).

In case it helps, the implementation for exploding geometries used by GeoPandas comes from PyGEOS / Shapely [here](https://github.com/shapely/shapely/blob/2.0a1/shapely/_geometry_helpers.pyx#L166) (in Cython).

We're able to pre-calculate the size of the resulting vector by summing the results from `GEOSGetNumGeometries_r` for each geometry in the input vector, then we fill each slot in the vector via `GEOSGetGeometryN_r`.  Performance-wise, I don't know how it would differ in Rust to build the vector as you iterate vs pre-allocate and fill, since the latter involves additional function calls but theoretically less moving of memory.

If I recall correctly, both the count and the individual geometry accessor return the next level down and does not recurse to the most granular geometry type.  So if you have a GeometryCollection(MultiPolygon(Polygon)), you'd get back a MultiPolygon.  Hopefully deeply-nested types are awkward, and a user can always detect that by inspecting the resulting types and loop until there are no more container types; seems just fine to not recurse in the implementation here either.

More important than the MultiIndex used by GeoPandas is simply getting back the index of the original geometries, so that you can join back to the original attributes.  So long as the individual (inner) geometries are returned in the same order as they occur in the input geometries, it is possible to calculate their inner index.  I'm not sure how often those inner indexes are used, but they seem less critical overall.

geopolars/src/geoseries.rs
edpft: Addressed in https://github.com/geopolars/geopolars/pull/111/commits/43f9edbe614720380e3d6b512595774a966885c5

geopolars/src/geoseries.rs
edpft: Addressed in https://github.com/geopolars/geopolars/pull/111/commits/28dd535bba9cf9d0126b21268240a60426a5143d

edpft:
To assess the options I wrote a benchmark exploding a series of 1 million two-point `MultiPoint`s into a series of 2 million `Point`s.

My initial implementation:

```sh
explode                 time:   [1.1605 s 1.1769 s 1.1977 s]
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe
```

With @kylebarron's  suggestion of iterating with the match arms and pushing directly to an `exploded_vector`:

```sh
explode                 time:   [1.0562 s 1.0753 s 1.0974 s]
                        change: [-10.774% -8.6362% -6.1199%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe
```

I also tried pre-allocating as, I think (apologies if I'm misinterpreting!), @brendan-ward is suggesting:

```diff
-        let mut exploded_vector = Vec::new();
+       let capacity = self.estimated_size();
+       let mut exploded_vector = Vec::with_capacity(capacity);
 ```

But it didn't didn't make a noticeable difference:

```sh
explode                 time:   [1.0825 s 1.1029 s 1.1260 s]
                        change: [-0.1609% +2.5676% +5.6027%] (p = 0.08 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  7 (7.00%) high mild
  5 (5.00%) high severe
```

I've since changed the benchmark to exploding a series of 45k two-point `MultiPoint`s into a series of 90k `Point`s so that `cargo bench` can complete 100 samples in 5s.

edpft:
> For the GeoDataFrame case, I'm curious if there's an existing polars function that would allow us to provide a list of indexes and have polars automatically duplicate rows. Maybe we could leverage join here somehow

> More important than the MultiIndex used by GeoPandas is simply getting back the index of the original geometries, so that you can join back to the original attributes. So long as the individual (inner) geometries are returned in the same order as they occur in the input geometries, it is possible to calculate their inner index. I'm not sure how often those inner indexes are used, but they seem less critical overall.

So, could a solution be for ``GeoSeries::explode` to return a `GeoDataFrame` of two series: the index of the original geometries and the exploded geometries? I can see the benefit when `GeoSeries::explode` is called in a future `GeoDataFrame::explode` method but is that necessary in the `GeoSeries` implementation?

I'm happy to take a swing at it but I feel it should be a separate issue / pull request.

kylebarron:
Thanks @brendan-ward for chiming in!

> In case it helps, the implementation for exploding geometries used by GeoPandas comes from PyGEOS / Shapely [here](https://github.com/shapely/shapely/blob/2.0a1/shapely/_geometry_helpers.pyx?rgh-link-date=2022-09-19T16%3A19%3A30Z#L166) (in Cython).

I figure when we get around to focusing on perf improvements we'll be looking at shapely code often 🙂.

-----


@edpft Awesome! Thanks for making a benchmark! Looks great and good to see that removing that extra vec allocation did make a difference! (I tried to figure out an example benchmarking setup previously but didn't get around to making many benches. btw is that output just `cargo bench`?)

I think @brendan-ward's suggestion here is more involved than just changing `Vec::with_capacity`. I think it would consist of two loops over the input data: one quick one just to count how big the output arrays should be. Then we can allocate arrays and just fill coordinates on the second loop. I do think this would likely be faster, but maybe not until we can move to an arrow-native geometry encoding (https://github.com/geopolars/geopolars/issues/31)

edpft:
Hi @kylebarron, yes, it's just `cargo bench`.

I thought @brendan-ward's suggestion might be more involved, but I was hoping I could approximate it by allocating a `Vec` with capacity. Is this something you'd like me to look at? I'm here for the learning more than anything, so I'm happy to keep working on this PR.

kylebarron:
> So, could a solution be for `GeoSeries::explode` to return a `GeoDataFrame` of two series: the index of the original geometries and the exploded geometries?

What I had in mind was returning a second `uint` column of indexes that mapped the row index of existing geometries into the row index of the exploded geometries (or the reverse, not sure which would be easier to work with).

But I don't know in general the best practices for returning two columns from a polars function. Like in pandas you can assign two columns onto an existing table by doing something like

```py
df = pd.DataFrame({'col1': [1, 2, 3], 'col2': [3, 4, 5]})
df[['col1_dup', 'col2_dup']] = df
```

and then `df` has 4 columns. Not sure what the usual way to do that in polars is. This is also why I got stuck on https://github.com/geopolars/geopolars/pull/11

> I'm happy to take a swing at it but I feel it should be a separate issue / pull request.

Yeah I agree, just saying it's useful to discuss more the end goal of what the function's API should be, which we can revisit in future PRs.

kylebarron:
> Is this something you'd like me to look at?

Honestly I don't think it's necessary at this stage. Most importantly your implementation is correct and plenty fast. My personal preference is to first focus on just getting things working, and then in the future we can come back and optimize performance

kylebarron (approved):

kylebarron:
Sometime in the future we'll also want to add this to the Python bindings, but it wouldn't hurt to discuss the explode API a bit more first

PR #117, opened 2022-09-25
WIP: multithreading support

kylebarron:
An absolutely massive performance improvement (at least for wall time), and shows that it's not hard to parallelize functions by row

```py
import geopolars as gpl
import polars as pl
gs = gpl.datasets.read_dataset('nybb').get_column('geometry')
# Concat 1000 times for a larger series
gs_larger = gpl.GeoSeries(pl.concat(1000* [gs]))

%time gs_larger.area
```

With `master`:
```
CPU times: user 9.94 s, sys: 35 ms, total: 9.98 s
Wall time: 9.98 s
```

With this branch:
```
CPU times: user 10.4 s, sys: 145 ms, total: 10.6 s
Wall time: 763 ms
```

With geopandas:
```
CPU times: user 719 ms, sys: 15.2 ms, total: 734 ms
Wall time: 750 ms
```

Lessons learned:

- Having to copy arrow data into a vec, parse that wkb `vec<u8>` into a geometry, then copy it back into a wkb is atrocious for performance. We already knew that but this puts real numbers to it
- This is still 14x slower than geopandas when you consider cpu time, but we have huge performance gains still possible by avoiding copies
- Can't wait to look at geoarrow encoding + traits

### DIFF
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -3,11 +3,11 @@ use crate::util::{iter_geom, parse_u8_series_to_geom};
 use geo::algorithm::affine_ops::AffineTransform;
 use geo::{map_coords::MapCoords, Geometry, Point};
 use geozero::{CoordDimensions, ToWkb};
+use polars::export::arrow;
 use polars::export::arrow::array::{
     Array, BinaryArray, BooleanArray, MutableBinaryArray, MutableBooleanArray,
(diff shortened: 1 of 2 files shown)


kylebarron:
Closing as stale.

In any case, we'll want to allow parallelization to happen from the Polars side, not row-by-row on our side.

PR #123, opened 2022-10-07
chore: Bump polars to 0.24

kylebarron:


.github/workflows/wasm.yml
```diff
@@ -1,10 +1,13 @@
 name: Wasm Build and Test
 
-on:
-  push:
-    branches:
-      - master
-  pull_request:
+# For now we turn off the WASM testing on PRs so that we can upgrade the rest of the code to polars
+# 0.24 without needing to touch our polars fork with wasm changes.
+on: workflow_dispatch
```
kylebarron, line 5: @stuartlynn turned off wasm CI (and it will fail to build for now) because your fork is still on 0.22.6

PR #137, opened 2022-10-15
Updated python linting/makefile

kylebarron:
try to update from latest polars


.github/workflows/lint-python.yml
```diff
@@ -0,0 +1,51 @@
+name: Lint Python
```
vinayakkulkarni, line 1: Maybe we can use an pre-built action - https://github.com/wearerequired/lint-action

?

kylebarron: I'm not familiar with that action; what does it help with? My original idea was to try and align our repo setup as much as possible with polars upstream

kylebarron:
Closing as stale

PR #147, opened 2022-10-19
try to build wheels with proj (macos testing)

kylebarron:
looks like this works (builds without error) when building against the proj installed from brew. I'll need to check that the import works correctly, and also need to check that the import works on M1 as well.


kylebarron:
The import works locally!

kylebarron:
We now have support thanks to several PRs: https://github.com/geopolars/geopolars/pull/170, https://github.com/geopolars/geopolars/pull/155, https://github.com/geopolars/geopolars/pull/163, https://github.com/geopolars/geopolars/pull/160

PR #155, opened 2022-10-29
Refactor wheel builds

kylebarron:
Notes:

- cibuildwheel doesn't allow us to have fine-grained control of the call to `maturin`. So e.g. we can't pass different features to maturin per OS. (maturin I don't think has a way to control it via env vars)
- Since cibuildwheel uses the native OS for windows and mac, could make sense to still use it for those platforms.
- Can cibuildwheel build abi3 wheels? abi3 wheels are useful to decrease the time to build wheels and generally simplify


kylebarron:
I should merge this without proj support and come back to that

PR #162, opened 2022-10-30
Wheels refactor (mac proj)

kylebarron:


kylebarron:
even when the wheels build correctly, it crashes at runtime because it can't find `proj.db`. I'll have to take a closer look at the pyproj setup

image

kylebarron:
Closing this as mac wheels appear to work with proj from homebrew

PR #166, opened 2022-11-04
Bundle proj data into the python wheels

nmandery:
This is an attempt build the python wheels with proj support by bundling the proj-data files into the wheel.

Bundling works so far, but to be able to tell proj about the location of the files ([using pythons pkg_resources](https://stackoverflow.com/questions/49773125/find-data-files-in-python3-wheel)) requires some API changes to `GeoSeries::to_crs`. In this method the transformation is constructed using the `Proj` struct, but to be able to pass in a custom data location, we would need to either pass in a `ProjBuilder` or create a custom options struct we can pass to that method. The first solution would leak proj types in the geopolars API, which currently has been avoided. The solution with the options struct may have some overheads by requiring to crate builders for each method call where the options are passed to. Thats probably no different from the current implementation, though.

@kylebarron Whats your opinion on this?

### DIFF
--- a/geopolars/Cargo.toml
+++ b/geopolars/Cargo.toml
@@ -40,7 +40,6 @@ thiserror = "1.0"
 
 proj = { version = "0.27.0", optional = true, features = [
   "geo-types",
-  "bundled_proj",
 ] }
 
 [dev-dependencies]
--- a/geopolars/Cargo.toml
+++ b/geopolars/Cargo.toml
@@ -23,6 +23,7 @@ default = [
 ]
 
 proj = ["dep:proj"]
+bundled_proj = ["proj", "proj/bundled_proj"]
 
 [dependencies]
 geozero = { version = "0.9.4", features = ["with-wkb"] }
--- a/geopolars/src/export.rs
+++ b/geopolars/src/export.rs
@@ -1,3 +1,2 @@
-
 #[cfg(feature = "proj")]
 pub use proj;
--- a/geopolars/src/export.rs
+++ b/geopolars/src/export.rs
@@ -1,0 +1,3 @@
+
+#[cfg(feature = "proj")]
+pub use proj;
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -9,6 +9,8 @@ use polars::export::arrow::array::{
     MutablePrimitiveArray, PrimitiveArray,
 };
 use polars::prelude::{PolarsError, Series};
+#[cfg(feature = "proj")]
+use proj::ProjBuilder;
 use std::convert::Into;
 
 pub type ArrayRef = Box<dyn Array>;
@@ -177,6 +179,16 @@ pub trait GeoSeries {
     // current CRS, but that would require polars to support extension types.
     #[cfg(feature = "proj")]
     fn to_crs(&self, from: &str, to: &str) -> Result<Series>;
+
+    // Note: Ideally we wouldn't have both `from` and `to` here, where the series would include the
+    // current CRS, but that would require polars to support extension types.
+    #[cfg(feature = "proj")]
+    fn to_crs_with_builder(
+        &self,
+        from: &str,
+        to: &str,
+        proj_builder: ProjBuilder,
+    ) -> Result<Series>;
 
     /// Returns a GeoSeries with each of the geometries translated by a fixed x and y amount
     ///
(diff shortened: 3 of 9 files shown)


kylebarron:
> we would need to either pass in a `ProjBuilder` or create a custom options struct we can pass to that method.

I don't really have a strong opinion here; I think I'd be ok with either. Do you have a preference?

geopolars/Cargo.toml
```diff
@@ -40,7 +40,6 @@ thiserror = "1.0"
 
 proj = { version = "0.27.0", optional = true, features = [
   "geo-types",
-  "bundled_proj",
```
kylebarron, line 43: Is this just for testing or do you prefer not using this flag?

nmandery: This was the only way I found to be able to find the proj_data directory - by using the systems proj installation. This also means we need to compile proj ourselves in the CI.

The outcome in the end changes little, though.[proj will still be statically linked to geopolars in the wheels](https://github.com/georust/proj/blob/d7185a7152db99ff34775e76064bc52c3ef45bd5/proj-sys/build.rs#L118).

nmandery:
> > we would need to either pass in a `ProjBuilder` or create a custom options struct we can pass to that method.
> 
> I don't really have a strong opinion here; I think I'd be ok with either. Do you have a preference?

Having types of dependencies in the public API forces the user of the crate to use the same version of the dependency as well. I this case I am not sure If it would be possible to safely use another proj version elsewhere in the same project without running into issues with the linked C++ library. So I suppose we can go with passing a `ProjBuilder`.

kylebarron:
> Having types of dependencies in the public API forces the user of the crate to use the same version of the dependency as well

I think polars generally solves this by having an `export` mod. Like we import `arrow2` via `polars::export::arrow`. So similar we'd be able to export proj as `geopolars::export::proj`
 
> So I suppose we can go with passing a `ProjBuilder`.

This is the solution that puts proj in our public API? Based on the context I thought you were going to suggest the opposite

geopolars/Cargo.toml
kylebarron: I'm ok with compiling PROJ ourselves in CI. We might still want to add a new top-level feature so that Rust consumers of `geopolars` would be able to opt in to bundled proj? I.e. if they don't have proj installed locally

nmandery: Good point. Added in d271a69b06a9d57e40042d8d9299618045aa97f5

nmandery:
Yes, passing `ProjBuilder` adds proj to the public API. I did not like that solution first, but now I think thats the most flexible one

nmandery:
With 8278fce the bundled proj_data files are loaded.

Running this test script 

```python
import geopandas.datasets
import geopolars as gpl

path = geopandas.datasets.get_path("naturalearth_cities")
gdf = gpl.read_file(path)
assert isinstance(gdf, gpl.GeoDataFrame)
gdf.geometry.to_crs("EPSG:4326", "EPSG:3857")
```

Results in these open attempts:

```bash
$ strace python test.py 2>&1 | grep proj.db | grep openat
openat(AT_FDCWD, "/home/nicodev/tmp/geopolars-proj-test/venv/lib/python3.10/site-packages/pyproj/proj_dir/share/proj/proj.db", O_RDONLY|O_NOFOLLOW|O_CLOEXEC) = 3
openat(AT_FDCWD, "/home/nicodev/tmp/geopolars-proj-test/venv/lib/python3.10/site-packages/fiona/proj_data/proj.db", O_RDONLY) = 4
openat(AT_FDCWD, "/home/nicodev/tmp/geopolars-proj-test/venv/lib/python3.10/site-packages/fiona/proj_data/proj.db", O_RDONLY|O_NOFOLLOW|O_CLOEXEC) = 4
openat(AT_FDCWD, "/home/nicodev/.local/share/proj/proj.db", O_RDONLY) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/home/nicodev/CLionProjects/geopolars/py-geopolars/target/debug/build/proj-sys-ab495d3c0adf4fd8/out/share/proj/proj.db", O_RDONLY) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/home/nicodev/tmp/geopolars-proj-test/venv/lib/python3.10/site-packages/geopolars/proj_data/proj.db", O_RDONLY) = 5
openat(AT_FDCWD, "/home/nicodev/tmp/geopolars-proj-test/venv/lib/python3.10/site-packages/geopolars/proj_data/proj.db", O_RDONLY|O_NOFOLLOW|O_CLOEXEC) = 5
```

The reason the proj.db in the target directory is attempted to be opened seems to be the construction of the `ProjBuilder`. I do not see how to remove that.

Now there is still to add the compilation of proj to the CI jobs ...

kylebarron:
Finished in https://github.com/geopolars/geopolars/pull/170

PR #170, opened 2022-11-14
bundle proj data into the python wheels (continued)

kylebarron:
continued from https://github.com/geopolars/geopolars/pull/166

cc @nmandery; I didn't appear to be able to push to your branch

### DIFF
--- a/geopolars/Cargo.toml
+++ b/geopolars/Cargo.toml
@@ -40,7 +40,6 @@ thiserror = "1.0"
 
 proj = { version = "0.27.0", optional = true, features = [
   "geo-types",
-  "bundled_proj",
 ] }
 
 [dev-dependencies]
--- a/geopolars/Cargo.toml
+++ b/geopolars/Cargo.toml
@@ -23,6 +23,7 @@ default = [
 ]
 
 proj = ["dep:proj"]
+bundled_proj = ["proj", "proj/bundled_proj"]
 
 [dependencies]
 geozero = { version = "0.9.4", features = ["with-wkb"] }
--- a/geopolars/src/export.rs
+++ b/geopolars/src/export.rs
@@ -1,3 +1,2 @@
-
 #[cfg(feature = "proj")]
 pub use proj;
--- a/geopolars/src/export.rs
+++ b/geopolars/src/export.rs
@@ -1,0 +1,3 @@
+
+#[cfg(feature = "proj")]
+pub use proj;
--- a/geopolars/src/export.rs
+++ b/geopolars/src/export.rs
@@ -1,2 +1,0 @@
-#[cfg(feature = "proj")]
-pub use proj;
--- a/geopolars/src/geoseries.rs
+++ b/geopolars/src/geoseries.rs
@@ -9,6 +9,8 @@ use polars::export::arrow::array::{
     MutablePrimitiveArray, PrimitiveArray,
 };
 use polars::prelude::{PolarsError, Series};
+#[cfg(feature = "proj")]
+use proj::ProjBuilder;
 use std::convert::Into;
 
 pub type ArrayRef = Box<dyn Array>;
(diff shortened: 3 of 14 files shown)


kylebarron:
@nmandery let me know what you think about these changes.

I changed from using `ProjBuilder` in the API directly to using a proj options struct because I _think_ we can't use `ProjBuilder` in a multithreaded context, because it doesn't even implement `clone`.

I.e. just to see if it would compile, I tried:
```rs
use proj::ProjBuilder;

let builder = ProjBuilder::default();

(0..1000).into_par_iter().for_each_with(
    || builder.proj_known_crs(from, to, None).unwrap(),
    |proj, x| {
        let proj = proj();
    },
);
```
There I get compiler errors of:
```
the trait bound `proj::ProjBuilder: std::clone::Clone` is not satisfied in `[closure@geopolars/src/geoseries.rs:817:13: 817:63]`
within `[closure@geopolars/src/geoseries.rs:817:13: 817:63]`, the trait `std::clone::Clone` is not implemented for `proj::ProjBuilder`rustc
```

However with `proj_options` it seems to work:
```rs
(0..1000).into_par_iter().for_each_with(
    || {
        proj_options
            .clone()
            .to_proj_builder()
            .unwrap()
            .proj_known_crs(from, to, None)
            .unwrap()
    },
    |proj, x| {
        let proj = proj();
    },
);
```

That appears to compile just fine.

I also tried to use the `thread_local` crate based on [this suggestion](https://stackoverflow.com/a/57452666), but that didn't work even with `proj_options`.
```rs
use thread_local::ThreadLocal;

let tl = ThreadLocal::new();

(0..1000).into_par_iter().for_each(|i| {
    let store = tl.get_or(|| {
        proj_options
            .clone()
            .to_proj_builder()?
            .proj_known_crs(from, to, None)?
    });
});
```
gives
```
`*mut proj_sys::PJconsts` cannot be sent between threads safely
within `proj::Proj`, the trait `std::marker::Send` is not implemented for `*mut proj_sys::PJconsts`
```

Not sure if I was doing all that correctly, but it would be nice to have seamless multithreading for reprojections.

kylebarron:
This appears to work on my local mac 🎉 

```py
In [1]: import geopolars as gpl

In [2]: gdf = gpl.datasets.read_dataset('naturalearth_cities')

In [3]: gs = gdf.get_column('geometry')

In [4]: test = gs.to_crs('epsg:4326', 'epsg:3857')

In [5]: test.to_geopandas()
Out[5]:
0        POINT (1386304.644 5146502.579)
1        POINT (1385011.523 5455558.181)
2        POINT (1059390.803 5963928.580)
3       POINT (3473167.790 -3056995.462)
4         POINT (682388.790 6379291.919)
                     ...
238    POINT (-4810350.913 -2620812.957)
239    POINT (-5190490.090 -2699486.457)
240    POINT (16832903.820 -4011543.664)
241      POINT (11560960.460 144168.711)
242     POINT (12710800.486 2548415.574)
Length: 243, dtype: geometry
```

kylebarron:
@nmandery I'm going to merge this so that I can tinker a bit more with the wheels, but if you have feedback on these changes I'd love to hear it still!

nmandery:
@kylebarron Thanks for finishing this - sadly I have not found the time to continue working on this in the recent days.

Looks good to me. I see you are doing the resolving of the proj_data directory now in python. 

The reasoning to use a `ProjOptions` struct to be able to construct multiple ProjBuilders for each thread makes perfect sense as neither `Proj` or `ProjBuilder` can be shared between threads. The compilation error is cause by the `ProjBuilder` returned by the closure not implementing `Send`, I suppose wrapping it in a `RefCell` would help. The ThreadLocal should ensure that only one thread is able to access it. I have not tried it, though ;)

kylebarron:
> I see you are doing the resolving of the proj_data directory now in python.

Yeah... seemed a little easier instead of using Rust to call into Python? Maybe I just instinctively reach for Python still 😄 

> I suppose wrapping it in a `RefCell` would help. The ThreadLocal should ensure that only one thread is able to access it

Cool; I'll get back to prototyping this sometime in the future! I'm not experienced at threading in Rust yet.

kylebarron:
Not sure if you saw but I also got wheel builds working with Proj for Mac, Mac M1, Linux, **and** Windows!

nmandery:
Thats great, so geopolars is getting closer to be releasable.

PR #187, opened 2023-01-07
geoarrow struct wip

kylebarron:
Progress on implementing a geoarrow struct encoding (see https://github.com/geoarrow/geoarrow/pull/26), which should be a lot faster than wkb


kylebarron:
A lot more work but I'll merge and iterate in a new branch

PR #202, opened 2023-01-17
generic scalar pointtrait

kylebarron:
ref https://github.com/georust/geo/discussions/838#discussioncomment-3797418


kylebarron:
stale and no longer relevant

PR #214, opened 2023-03-20
add `points_from_xy` to `GeoSeries` class

tsutterley:
Parallels function of same name in GeoPandas for creating a GeoSeries from in-memory arrays

### DIFF
--- a/py-geopolars/python/geopolars/internals/geoseries.py
+++ b/py-geopolars/python/geopolars/internals/geoseries.py
@@ -206,27 +206,25 @@ class GeoSeries(pl.Series):
         raise ValueError()
 
     @classmethod
-    def points_from_xy(cls, x, y,
-                        z: Iterable | pl.Series | None = None,
-                        crs: str | pyproj.CRS | None = None
-                        ):
+    def points_from_xy(
+        cls,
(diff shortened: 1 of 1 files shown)


py-geopolars/python/geopolars/internals/geoseries.py
```diff
@@ -205,6 +205,29 @@ def geoarrow_coords_to_numpy(struct_array: pyarrow.StructArray):
 
         raise ValueError()
 
+    @classmethod
+    def points_from_xy(cls, x, y,
```
kylebarron, line 209: I think you also want the type hints on `x` and `y`? Note that `Iterable | pl.Series` is a strict subset of what you can pass into `pyarrow.StructArray`; it also accepts numpy arrays and pyarrow arrays I think

kylebarron:
Thanks for the PR! I'm not exactly sure at a glance why CI is now failing. Regardless I think you might need to apply `black`

tsutterley:
Removing z makes things simpler :)
applied `black` formatting so we'll see how it goes with CI
happy to contribute!

kylebarron:
Closing as stale. Now that Polars supports extension types, this will hopefully be unblocked soon

PR #240, opened 2024-08-18
Revamp Python API as pure-Python wrapper around polars, arro3, and geoarrow-rust-core

kylebarron:
After a bit of exploration on this, I don't think it's worth going forward with this approach. The hope was that we could provide a relatively simple pure-python wrapper that adds extra metadata on the python side, but I don't think that will be feasible because dealing with a subclass of polars.DataFrame will be an endless battle:

- polars explicitly [does not support subclassing of `DataFrame` and `Series`](https://github.com/pola-rs/polars/issues/2846#issuecomment-1711799869). So if you add extra metadata onto `GeoDataFrame`, that metadata will get lost on every non-overridden method, because those will return a `DataFrame`. So we'd have to override every single polars method in the subclass to return a GeoDataFrame, and also keep track of which methods might adjust the schema. 
- We need some sort of way to persist metadata in polars. Dataframe-level metadata is still incomplete because methods like `df.rename` will change the content of the columns without changing the top-level metadata.
- Really the only viable option (in my opinion) is to have extension types supported by polars, and it seems they have no interest in doing that. See https://github.com/pola-rs/polars/issues/9112 and https://github.com/pola-rs/polars/issues/7288#issuecomment-1957537542


Andre-Medina:
Great to see your development in [GeoArrow](https://github.com/geoarrow/geoarrow-rs) and support for arrow extention types in [Polars](https://github.com/pola-rs/polars) is on the roadmap. So `GeoPolars` will exist someday :)

In the meantime, packages like [polars-st](`https://github.com/Oreilles/polars-st`) have popped up as self labelled tempoarry solutions to the spacial data in polars problem.

Were there any intentions to standardizing an API for `GeoPolars` itself to make a transition from an alternate solution easier? I am guessing it would be hard to plan an API while `GeoArrow` and extension types are still under development.

kylebarron:
I don't think it's possible to standardize an API now. It depends partially on what the polars integration APIs look like if and when they enable the features we need.

kylebarron:
Closing as stale now that Polars supports extension types

PR #258, opened 2025-12-10
feat: Implement mapping from GeoArrow extension types to Polars extension types

kylebarron:
### Change list

- Create `geopolars-extension` crate to handle definitions of Polars extension types and the act of registering them with Polars
- Create Polars extension types as wrappers around upstream types in `geoarrow-schema`
- Create factory for dynamically creating extension types.

Future work: 

- allow extension types to be fallible: https://github.com/pola-rs/geopolars/issues/260


geopolars/geopolars-extension/src/factory.rs
```diff
@@ -0,0 +1,74 @@
...
+/// A factory for creating GeoArrow extension types in Polars.
+pub struct GeoArrowExtensionTypeFactory;
+
+impl ExtensionTypeFactory for GeoArrowExtensionTypeFactory {
+    fn create_type_instance(
+        &self,
+        name: &str,
+        storage: &DataType,
+        metadata: Option<&str>,
+    ) -> Box<dyn ExtensionTypeImpl> {
```
kylebarron, line 22: @orlp I was surprised that this function is infallible. How should implementors behave when the `storage` type is incompatible with the declared extension type for the given `name`?

This corresponds to [`arrow_schema::ExtensionType::try_new`](https://docs.rs/arrow-schema/latest/arrow_schema/extension/trait.ExtensionType.html#tymethod.try_new), where they allow an error in case

> This should return an error if the given data type is not supported by this extension type.

geopolars/geopolars-extension/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use arrow_schema::extension::ExtensionType;
+use polars::error::PolarsResult;
+use polars_core::datatypes::extension::register_extension_type;
+
+use crate::factory::GeoArrowExtensionTypeFactory;
+
+/// Register all GeoArrow extension types into the Polars extension registry.
+pub fn register_all_extensions() -> PolarsResult<()> {
+    let factory = Arc::new(GeoArrowExtensionTypeFactory);
```
kylebarron, line 14: We could create a separate factory function for each geometry type, but for simplicity I think it's fine to have a single factory for all types

geopolars/geopolars-extension/src/geoarrow.rs
```diff
@@ -0,0 +1,120 @@
...
+    MultiPolygonType
+);
+define_basic_type!(
+    /// A GeoPolars GeometryCollection extension type.
+    GeometryCollectionType
+);
+define_basic_type!(
+    /// A GeoPolars Geometry extension type.
+    GeometryType
+);
```
kylebarron, line 108: These two types are implemented with Arrow unions, so they're currently incompatible with Polars. They're still defined for completeness, but I assume these code paths will never be reached, because Polars will error at some prior point when seeing union types.

geopolars/geopolars-extension/src/geoarrow.rs
```diff
@@ -0,0 +1,120 @@
...
+
+                self == other
+            }
+
+            fn dyn_hash(&self) -> u64 {
+                PlFixedStateQuality::default().hash_one(self)
+            }
+
+            fn dyn_display(&self) -> Cow<'_, str> {
+                Cow::Owned(format!("{:?}", self.0))
```
kylebarron, line 55: In the future we can better define a `Display` that is improved over `Debug`

geopolars/geopolars-extension/src/factory.rs
orlp: For the moment there is no way to register a type-check for storage type compatibility. Like any other overridable behavior there were unresolved questions about multiple extensions registering incompatible type-checks. The rest of the Polars code base also was not yet ready for the concept of "give me the data type of this column" to be fallible (which would be the case when reading from Parquet with an extension type with a fallible constructor).

orlp: The behavior should just be to construct some valid representation. Do your type checking for the correct storage type in your functions handling the data, at least for now.

kylebarron: Ok, perhaps I'll change this to be
```rs
struct PointType(GeoArrowResult<geoarrow_schema::PointType>);
```
but I'll leave that for a future PR

